PR #11, opened 2026-04-05
Make --host understood on subcommands

### LINKED ISSUE #10: --host is an unrecognized argument
``` shell
$ canasta create --host wikiapiary.com --id wiki-prod --wiki main
usage: canasta [-h] [--host HOST] [--verbose]  ...
canasta: error: unrecognized arguments: --host wikiapiary.com
```

hexmode:
Fixes #10

### DIFF
--- a/canasta.py
+++ b/canasta.py
@@ -393,6 +393,7 @@ def main():
             pre_cmd.append(arg)
             skip_next = False
         elif arg in global_value_flags:
+            # --host before command: add to pre_cmd and skip value
             pre_cmd.append(arg)
             skip_next = True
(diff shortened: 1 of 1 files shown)


cicalese:
Superseded by 22d522c. The PR's approach has bugs: it doesn't respect the '--' passthrough separator (so a --host inside passthrough args gets incorrectly extracted), its value check only rejects other global flags as values (so 'canasta create --host --wait 30' would extract '--host --wait'), and it doesn't handle the --host=value form. The replacement commit fixes those cases and adds 9 unit tests.

PR #117, opened 2026-04-14
Add some debug messages

### LINKED ISSUE #98: Canasta sits too long without feedback
Something is going on, but nothing is displayed.  It would be nice to have some feedback after an initial implementation:  

```shell
$ canasta -H root@meta.wikiapiary.com create -i wikiapiary -w wikiapiary -n meta.wikiapiary.com
[ ... way too much time ... ]
Generated root database password (will be saved to .env)
Generated admin password.
Successfully written to wikis.yaml
Creating Canasta instance 'wikiapiary'...
[ ... way too much time ... ]
```

### LINKED ISSUE #99: First run of canasta fails
```shell
$ canasta -H root@meta.wikiapiary.com creatry.com
Generated root database password (will be saved to .env)
Generated admin password.
Successfully written to wikis.yaml
Creating Canasta instance 'wikiapiary'...
Error: non-zero return code
unknown shorthand flag: 'f' in -f
See 'docker --help'.

Usage:  docker [OPTIONS] COMMAND

A self-sufficient runtime for containers

Common Commands:
  run         Create and run a new container from an image
  exec        Execute a command in a running container
  ps          List containers
  build       Build an image from a Dockerfile
  pull        Download an image from a registry
  push        Upload an image to a registry
  images      List images
  login       Log in to a registry
  logout      Log out from a registry
  search      Search Docker Hub for images
  version     Show the Docker version information
  info        Display system-wide information

Management Commands:
  builder     Manage builds
  buildx*     Docker Buildx
  container   Manage containers
  context     Manage contexts
  image       Manage images
  manifest    Manage Docker image manifests and manifest lists
  network     Manage networks
  plugin      Manage plugins
  system      Manage Docker
  trust       Manage trust on Docker images
  volume      Manage volumes

Swarm Commands:
  swarm       Manage Swarm

Commands:
  attach      Attach local standard input, output, and error streams to a running container
  commit      Create a new image from a container's changes
  cp          Copy files/folders between a container and the local filesystem
  create      Create a new container
  diff        Inspect changes to files or directories on a container's filesystem
  events      Get real time events from the server
  export      Export a container's filesystem as a tar archive
  history     Show the history of an image
  import      Import the contents from a tarball to create a filesystem image
  inspect     Return low-level information on Docker objects
  kill        Kill one or more running containers
  load        Load an image from a tar archive or STDIN
  logs        Fetch the logs of a container
  pause       Pause all processes within one or more containers
  port        List port mappings or a specific mapping for the container
  rename      Rename a container
  restart     Restart one or more containers
  rm          Remove one or more containers
  rmi         Remove one or more images
  save        Save one or more images to a tar archive (streamed to STDOUT by default)
  start       Start one or more stopped containers
  stats       Display a live stream of container(s) resource usage statistics
  stop        Stop one or more running containers
  tag         Create a tag TARGET_IMAGE that refers to SOURCE_IMAGE
  top         Display the running processes of a container
  unpause     Unpause all processes within one or more containers
  update      Update configuration of one or more containers
  wait        Block until one or more containers stop, then print their exit codes

Global Options:
      --config string      Location of client config files (default "/root/.docker")
  -c, --context string     Name of the context to use to connect to the daemon (overrides DOCKER_HOST env var and default context set with "docker context use")
  -D, --debug              Enable debug mode
  -H, --host list          Daemon socket to connect to
  -l, --log-level string   Set the logging level ("debug", "info", "warn", "error", "fatal") (default "info")
      --tls                Use TLS; implied by --tlsverify
      --tlscacert string   Trust certs signed only by this CA (default "/root/.docker/ca.pem")
      --tlscert string     Path to TLS certificate file (default "/root/.docker/cert.pem")
      --tlskey string      Path to TLS key file (default "/root/.docker/key.pem")
      --tlsverify          Use TLS and verify the remote
  -v, --version            Print version information and quit

Run 'docker COMMAND --help' for more information on a command.

For more help on how to use Docker, head to https://docs.docker.com/go/guides/
Error: Instance creation failed. Files cleaned up.
```

hexmode:
And make sure docker compose is installed

Fixes #99
Fixes #98

### DIFF
--- a/roles/create/tasks/main.yml
+++ b/roles/create/tasks/main.yml
@@ -355,41 +355,38 @@
     mode: "0755"
   loop: "{{ _created_wikis.wiki_ids }}"
 
+- when: wiki_settings is defined and wiki_settings != ''
+  ansible.builtin.debug:
+    msg: "Copying per-wiki Settings.php"
+
 - name: Copy per-wiki Settings.php if provided
   when: wiki_settings is defined and wiki_settings != ''
-  block:
-    - name: Print it out
-      ansible.builtin.debug:
-        msg: "Copying per-wiki Settings.php"
-    - name: do it
-      ansible.builtin.copy:
-        src: "{{ wiki_settings }}"
-        dest: "{{ instance_path }}/config/settings/wikis/{{ wiki }}/Settings.php"
-        mode: "0644"
+  ansible.builtin.copy:
+    src: "{{ wiki_settings }}"
+    dest: "{{ instance_path }}/config/settings/wikis/{{ wiki }}/Settings.php"
+    mode: "0644"
+
+- when: global_settings is defined and global_settings != ''
+  ansible.builtin.debug:
+    msg: "Copying global settings"
 
 - name: Copy global settings if provided
   when: global_settings is defined and global_settings != ''
-  block:
-    - name: Print it out
-      ansible.builtin.debug:
-        msg: "Copying global settings"
-    - name: do it
-      ansible.builtin.copy:
-        src: "{{ global_settings }}"
-        dest: "{{ instance_path }}/config/settings/global/{{ global_settings | basename }}"
-        mode: "0644"
+  ansible.builtin.copy:
+    src: "{{ global_settings }}"
+    dest: "{{ instance_path }}/config/settings/global/{{ global_settings | basename }}"
+    mode: "0644"
+
+- when: composer is defined and composer != ''
+  ansible.builtin.debug:
+    msg: "Copying composer.local.json"
 
 - name: Copy composer.local.json if provided
   when: composer is defined and composer != ''
-  block:
-    - name: Print it out
-      ansible.builtin.debug:
-        msg: "Copying composer.local.json"
-    - name: do it
-      ansible.builtin.copy:
-        src: "{{ composer }}"
-        dest: "{{ instance_path }}/config/composer.local.json"
-        mode: "0644"
+  ansible.builtin.copy:
+    src: "{{ composer }}"
+    dest: "{{ instance_path }}/config/composer.local.json"
+    mode: "0644"
 
 # --- Step 9b: Disable auto-HTTPS when requested or when on K8s ---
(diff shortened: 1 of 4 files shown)


cicalese (commented):
Review from Claude: Thanks for this PR! The long wait during create is a real pain point and adding progress messages is the right approach.

A few suggestions:

**Task naming:** Some task names like "Print it out" and "do it" should be more descriptive — these show up in verbose output and logs.

**Lighter approach:** Rather than adding a debug message before every individual task, consider a few progress messages at major milestones only:
- Connecting to host
- Creating instance directory
- Pulling Docker images (this is the longest wait)
- Installing MediaWiki
- Done

The `canasta_minimal` callback plugin already shows `debug` messages to the user, so these will appear in non-verbose mode too — which means too many messages will be noisy.

**`noisy_create_dir.yml`:** Using `include_tasks` for each directory creation adds overhead. A single debug message before the loop would achieve the same effect without the extra file.

**Block wrapping:** Wrapping simple `copy` tasks in blocks just to add a debug message adds a lot of YAML complexity. A debug task before the existing task (without a block) is simpler.

**Unrelated changes:** The quote style changes (`"0644"` → `'0644'`) should be in a separate PR to keep the diff focused.

**Docker Compose check:** The `docker compose version` check is a good addition — that's actually part of #118 though. If included here, it should keep the original Docker failure message as a fallback.

hexmode:
> **`noisy_create_dir.yml`**: Using `include_tasks` for each directory creation adds overhead. A single debug message before the loop would achieve the same effect without the extra file.

I agree that is simpler, but it also doesn't work.  Each directory creation takes some time and if you only print out one message at the beginning it ends up sitting for a long time.

hexmode:
> **Block wrapping**: Wrapping simple copy tasks in blocks just to add a debug message adds a lot of YAML complexity. A debug task before the existing task (without a block) is simpler.

I agree, but it was a tradeoff between that and DRY.  I'll change it to the wet version.

PR #118, opened 2026-04-14
Install docker

### LINKED ISSUE #101: canasta should install docker if it needs to
```shell
$ canasta -H root@meta.wikiapiary.com create -i wikiapiary -w wikiapiary -n meta.wikiapiary.com
Error: Docker is not available on this host. Run 'canasta doctor' to check all d
Error: Instance creation failed. Files cleaned up.
```

Options include  

-   `apt install -y docker.io docker-compose`
-   `dnf install -y docker-cli containerd docker-compose`
-   etc

hexmode:
Fixes #101

### DIFF
--- a/roles/create/tasks/main.yml
+++ b/roles/create/tasks/main.yml
@@ -17,14 +17,18 @@
   ignore_errors: true  # OK if Docker not installed
   when: orchestrator | default('compose') == 'compose'
 
-- name: Fail if Docker not available
-  ansible.builtin.fail:
-    msg: >-
-      Docker is not available on this host.
-      Run 'canasta doctor' to check all dependencies.
-  when:
-    - orchestrator | default('compose') == 'compose'
-    - _docker_check.rc | default(0) != 0
+- name: Make sure docker compose is installed
+  ansible.builtin.command:
+    cmd: docker compose version
+  register: _docker_compose_check
+  changed_when: false
+  ignore_errors: true
+  when: orchestrator | default('compose') == 'compose'
+
+- name: Install Docker and/or Docker Compose
+  ansible.builtin.include_role:
+    name: docker
+    tasks_from: install_docker.yml
 
 # Compose-only host memory preflight. The K8s path has its own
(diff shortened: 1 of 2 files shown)


cicalese (commented):
Review from Claude: Thanks for tackling this — auto-installing Docker is a great quality-of-life improvement.

A few things that need addressing:

**Conditional install:** The PR removes the "Docker not available" error entirely. Instead, the install should only run when Docker is missing. If the install fails, there should still be a clear error message. Something like:
```yaml
- name: Install Docker if not available
  when:
    - _docker_check.rc | default(0) != 0
    - orchestrator | default('compose') == 'compose'
  ...
```

**Ubuntu support:** The apt repository URL hardcodes `debian`:
```
https://download.docker.com/linux/debian {{ ansible_distribution_release }} stable
```
This won't work on Ubuntu. Use `{{ ansible_distribution | lower }}` instead of `debian`.

**Architecture:** `arch=amd64` is hardcoded. For ARM hosts (e.g. AWS Graviton), this needs to be `{{ ansible_architecture | replace('x86_64', 'amd64') | replace('aarch64', 'arm64') }}` or similar.

**Docker group:** After installing Docker, the Ansible user needs to be added to the `docker` group to run Docker commands without sudo. Without this, subsequent commands will fail.

**Debug messages:** The "Debug" messages with generic text should either have useful content or be removed before merge.

**Already installed check:** If Docker and Docker Compose are both already installed, the role should skip entirely rather than re-running the apt/dnf tasks.

**Docker Compose check from #117:** The `docker compose version` check that's also in #117 should live in one PR or the other, not both.

cicalese:
Superseded by PR #197 which adds a dedicated `canasta install` command: `canasta install docker`, `canasta install k3s`, `canasta install git-crypt` (or all at once: `canasta install docker k3s git-crypt`). This separates dependency installation from `canasta create`, avoiding the surprise side-effect of auto-installing Docker during instance creation. The Docker installer in #197 is based on the work here, cleaned up with architecture detection, daemon start, user group handling, and platform guard (macOS/Windows get a helpful error).

PR #127, opened 2026-04-15
Update message for already existing instance

### LINKED ISSUE #104: Confusing error message
If I run `create` after the system has already been created, I see `Error: Instance 'wikiapiary' already exists at /root/wikiapiary` but (because I've re-deployed a new VM) there is no `/root/wikiapiary` on the host.  

If I `rm -rf ~/.config/canasta`, the command works, which causes me to think that the message should be `Error: Instance 'wikiapiary' already exists in the local configuration` or similar.

hexmode:
This changes:
```
Error: Instance 'wikiapiary' already exists at /root/wikiapiary
```
to:
```
Error: Instance 'wikiapiary' already exists in local config
Host: root@meta.wikiapiary.com
Path: /root/wikiapiary
```
Fixes #104

### DIFF
--- a/playbooks/create.yml
+++ b/playbooks/create.yml
@@ -18,7 +18,7 @@
 
 - name: Fail early if instance already exists
   ansible.builtin.fail:
-    msg: "Instance '{{ id }}' already exists at {{ _pre_check.instance.path }}"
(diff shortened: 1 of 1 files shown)


cicalese:
Rebased onto current main — dropped the stale CANASTA_VERSION and multi-node.md changes (already merged separately). Also fixed a variable name bug: the error message referenced `_existing.instance.path` but the register is `_pre_check`. Ready to merge.

PR #145, opened 2026-04-15
version: show 'unknown' instead of blank when git fallback returns empty

### LINKED ISSUE #133: canasta version: empty commit/built when git fallback fails (ownership mismatch)
## Problem

`canasta version` prints empty `commit` and `built` fields when the install directory's git fallback fails:

```
Canasta CLI v4.0.0 (native, commit , built )
```

## Repro

A common way to hit this is installing as root but running as a non-root user — modern git rejects the operation with "dubious ownership in repository":

```bash
sudo git clone https://github.com/CanastaWiki/Canasta-Ansible.git /opt/canasta-ansible
# ... set up venv as root, symlink canasta into /usr/local/bin ...
canasta version   # run as a non-root user
# Canasta CLI v4.0.0 (native, commit , built )
```

The `git rev-parse --short HEAD` call in `playbooks/version.yml` fails (root-owned `.git`, non-root caller), `ignore_errors: true` swallows the failure, and `_git_commit.stdout` ends up as an empty string.

## Root cause

`playbooks/version.yml` line 74:

```yaml
else (_git_commit.stdout | default('unknown') | trim)
```

Jinja2's `default('unknown')` only triggers on **undefined** values, not empty strings. When git fails under `ignore_errors: true`, `_git_commit.stdout` is `""` (defined but empty), so `default` doesn't fire and the output is blank.

## Proposed fix

Use the boolean form of `default`, which treats empty strings as falsy:

```yaml
else (_git_commit.stdout | default('unknown', true) | trim)
```

Same fix for `_version_date`.

Optional follow-up: when running as a different user than the install owner, surface the dubious-ownership message somewhere visible (e.g., during `canasta doctor`) so users understand why git operations might silently fail. Not strictly part of this bug.

cicalese:
## Summary
- Use Jinja's boolean `default` so empty-string results from the git fallback become 'unknown' instead of rendering blank.

## Test plan
- [x] Reproduced the blank output in a root-owned `/opt/canasta-ansible` clone run as a non-root user — git fails with dubious-ownership, stdout ends up empty. After the fix, `canasta version` renders "commit unknown, built unknown".

Fixes #133.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,7 @@ admin-password_*
 values-configdata.yaml
 inventory/hosts.yml
 .DS_Store
+
+# Version metadata captured at install time (Makefile: build-info)
+BUILD_COMMIT
+BUILD_DATE
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: test-unit test-integration test lint docs validate audit-coverage clean
+.PHONY: test-unit test-integration test lint docs validate audit-coverage clean build-info
 
 VENV := .venv
 PYTHON := $(VENV)/bin/python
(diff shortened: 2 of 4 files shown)


cicalese:
Added a second commit addressing the root cause: new 'make build-info' target captures BUILD_COMMIT/BUILD_DATE at install time, bypassing the runtime git fallback entirely. README Option 2 now includes `sudo make build-info`. The Jinja fallback remains as defense in depth for setups that skip the target.

PR #195, opened 2026-04-17
Improve 'instance already exists' error message (rebased #127)

### LINKED ISSUE #104: Confusing error message

cicalese:
## Summary

Rebased version of #127 (hexmode). Cherry-picked the two create.yml commits onto current main, dropping the stale CANASTA_VERSION and multi-node.md changes that were already merged.

Improves the "instance already exists" error message to show host, path, and a delete hint:

```
Instance 'X' already exists in the registry
(path: /root/X, host: root@example.com).
Run 'canasta delete --id X --yes' to remove it first.
```

Also fixed a variable name bug in the original (`_existing` → `_pre_check` to match the register).

Fixes #104. Supersedes #127.

### DIFF
--- a/playbooks/create.yml
+++ b/playbooks/create.yml
@@ -18,7 +18,7 @@
 
 - name: Fail early if instance already exists
   ansible.builtin.fail:
-    msg: "Instance '{{ id }}' already exists at {{ _pre_check.instance.path }}"
+    msg: "Instance '{{ id }}' already exists in local config\nHost: {{ _pre_check.instance.host }}\nPath: {{ _pre_check.instance.path }}"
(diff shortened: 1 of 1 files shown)


cicalese:
Closed and rebased in #127 instead.

PR #217, opened 2026-04-18
Map standard AWS env vars to short names in compose template

### LINKED ISSUE #137: Deduplicate AWS credential env vars: AWS_ACCESS_KEY / AWS_SECRET_KEY vs AWS_ACCESS_KEY_ID / AWS_SECRET_ACCESS_KEY
## Problem

The Canasta `.env` carries two pairs of AWS credentials for what are effectively the same values:

- **Standard SDK names** (used by Restic, AWS CLI, boto, most tooling):
  - `AWS_ACCESS_KEY_ID`
  - `AWS_SECRET_ACCESS_KEY`

- **Short non-standard names** (used by the MediaWiki AWS extension's `LocalSettings.php`):
  - `AWS_ACCESS_KEY`
  - `AWS_SECRET_KEY`

In practice both pairs hold the same values. Carrying two copies is fragile: users who see them and think "these look like duplicates" delete one pair, and something silently breaks — Restic fails to talk to S3 (needs the standard names), or the AWS extension fails to store images (needs the short names), or the docker-compose file emits warnings because the short names aren't set.

## Repro

Delete `AWS_ACCESS_KEY` and `AWS_SECRET_KEY` from `.env`:

```
docker compose ... warning: The "AWS_ACCESS_KEY" variable is not set. Defaulting to a blank string.
```

Then image uploads break because the AWS extension never sees credentials.

## Proposed fix

Converge on a single canonical pair (the standard `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY`), and either:

1. Update the Canasta image's AWS extension config to read the standard names.
2. Map standard names to short names at container start (e.g., `AWS_ACCESS_KEY=${AWS_ACCESS_KEY_ID}` in the entrypoint or compose env section).

Option 1 is the cleanest long-term fix; option 2 is a safe interim.

## Related

This surfaced while debugging a restore of the PGE wiki to a fresh test host — a user saw `AWS_ACCESS_KEY=x` and `AWS_ACCESS_KEY_ID=x` next to each other and removed what they thought was a duplicate.

cicalese:
## Problem
Users had to maintain two pairs of identical AWS credentials in `.env`:
- `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` (standard, used by Restic/AWS CLI)
- `AWS_ACCESS_KEY` / `AWS_SECRET_KEY` (non-standard, used by MW AWS extension)

Deleting either pair broke something silently.

## Fix
Map standard → short in the compose template's web environment:
```yaml
- AWS_ACCESS_KEY=${AWS_ACCESS_KEY_ID:-}
- AWS_SECRET_KEY=${AWS_SECRET_ACCESS_KEY:-}
```

Users only need the standard pair in `.env`. Also passes `AWS_REGION` and `AWS_BUCKET_NAME`.

## Backward compat
Existing instances with both pairs in `.env`: the docker-compose.override.yml (which has the short names explicitly) takes precedence. No change in behavior.

Fixes #137.

### DIFF
--- a/roles/orchestrator/files/compose/docker-compose.yml
+++ b/roles/orchestrator/files/compose/docker-compose.yml
@@ -62,6 +62,13 @@ services:
       - CANASTA_ENABLE_WIKI_DIRECTORY=${CANASTA_ENABLE_WIKI_DIRECTORY:-false}
       - CANASTA_ENABLE_VERY_SHORT_URLS=${CANASTA_ENABLE_VERY_SHORT_URLS:-false}
       - MW_SITEMAP_PAUSE_DAYS=${MW_SITEMAP_PAUSE_DAYS:-1}
+      # AWS: map standard SDK names to the short names the MW AWS
+      # extension reads. Users only need to set the standard pair
+      # (AWS_ACCESS_KEY_ID / AWS_SECRET_ACCESS_KEY) in .env.
+      - AWS_ACCESS_KEY=${AWS_ACCESS_KEY_ID:-}
+      - AWS_SECRET_KEY=${AWS_SECRET_ACCESS_KEY:-}
+      - AWS_REGION=${AWS_REGION:-}
+      - AWS_BUCKET_NAME=${AWS_BUCKET_NAME:-}
     volumes:
       - ./extensions:/var/www/mediawiki/w/user-extensions
       - ./skins:/var/www/mediawiki/w/user-skins


cicalese:
Wrong approach — AWS vars belong in per-instance docker-compose.override.yml, not the core template.

PR #219, opened 2026-04-18
Use mitogen

### LINKED ISSUE #218: Fast path “native”: Use mitogen
I know you're looking at another method but [mitogen](https://mitogen.networkgenomics.com/ansible_detailed.html) is a way to improve the speed of ansible without writing any other code.  It really speeds things up makes my debugging statements superfluous.

hexmode:
Fixes #218

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,6 +43,9 @@ COPY . .
 # Make wrapper executable
 RUN chmod +x canasta-native
 
+# Create symlink for mitogen strategy plugins (dynamic path resolution)
+RUN ln -s $(python -c "import site; print(site.getsitepackages()[0])")/ansible_mitogen/plugins/strategy /opt/canasta-ansible/.mitogen-strategy-plugins
+
 # Build metadata (injected by CI)
 ARG BUILD_COMMIT=unknown
 ARG BUILD_DATE=unknown
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,14 +1,8 @@
 [defaults]
-# Optimized fact collection
-gathering = smart
-# Cache to disk
-fact_caching = jsonfile
-fact_caching_connection = /tmp/ans_cache
-host_key_checking = False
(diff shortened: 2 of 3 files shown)


cicalese:
Tested mitogen with the correct version (0.3.46) on ansible-core 2.20.4 — it works. I pushed a fix to this branch:

- `mitogen>=0.3.33` in requirements.txt (minimum for ansible-core 2.20)
- `strategy = mitogen_linear` (not `mitogen`)
- Restored `StrictHostKeyChecking=accept-new` (needed for first-time host connections)
- Kept `gathering = explicit` (our playbooks don't use gathered facts)
- Removed fact caching and `host_key_checking = False`
- Kept `pipelining = True` (good)

### Benchmark: `canasta list` with 3 remote hosts

| Approach | Warm avg |
|---|---|
| Ansible (original) | ~9.0s |
| **Ansible + Mitogen** | **~7.4s** |
| Direct command bypass (#220) | ~0.9s |

Mitogen provides a modest ~18% improvement over vanilla Ansible, which is worth having for commands that stay on the Ansible path (create, upgrade, backup, etc.). For simple read-only commands, the direct bypass in #220 is ~8x faster than mitogen because it eliminates both Ansible startup overhead and sequential task execution.

### Deprecation concern

Ansible 2.19 deprecated third-party strategy plugins ([ansible/ansible#84725](https://github.com/ansible/ansible/issues/84725)). Mitogen's entire integration is a strategy plugin. Per the standard 4-release deprecation cycle, this mechanism would be removed around ansible-core 2.23. The mitogen maintainer has acknowledged this in [mitogen#1278](https://github.com/mitogen-hq/mitogen/issues/1278) but no alternative integration approach has been designed yet.

This is still worth merging for the near-term benefit to complex commands, but we should track the deprecation timeline.

cicalese:
Thank you for your contribution!

PR #449, opened 2026-05-01
Replace ansible_connection dance with delegate_to: localhost (closes #426)

### LINKED ISSUE #426: resolve_instance and delete/main.yml: replace ansible_connection save/switch/restore with delegate_to: localhost
## Summary

Two places in the codebase manipulate `ansible_host` / `ansible_connection` / `ansible_user` via `set_fact` to "switch to local" before calling registry tasks, then restore the original values afterward. This is fragile (the restore step has to mirror the save step exactly, and any failure between save and restore leaves the play in a weird state) and inconsistent with the rest of the codebase, which uses `delegate_to: localhost` + a `vars: { ansible_connection: local, ansible_python_interpreter: ... }` block on the registry task itself.

## Locations

- `roles/common/tasks/resolve_instance.yml:24-29` — save block
- `roles/common/tasks/resolve_instance.yml:91-96` — restore block (with three `canasta_registry` calls between them at lines 43, 75, 86)
- `roles/delete/tasks/main.yml:226-247` — same pattern, copy
- `roles/delete/tasks/main.yml:266-287` — same pattern, copy

## Proposed fix

Replace the save/switch/restore dance with `delegate_to: localhost` on each of the `canasta_registry` calls, matching the pattern used in `roles/create/tasks/main.yml`, `roles/upgrade/`, `roles/devmode/`, and elsewhere. Drop the now-unneeded save and restore tasks.

## Why

- Removes a class of bugs (forgotten restore on early failure, partial restore if the save block grows new fields).
- Makes the registry-call pattern uniform across the repo — easier for new contributors to follow.
- Fewer lines.

## Acceptance

Existing tests should continue to pass. `tests/unit/` registry tests are the primary signal. Worth running `make test-unit` and the relevant Molecule scenarios.

cicalese:
## Summary

Replace the brittle `ansible_connection` save/switch/restore dance in `resolve_instance.yml` and `delete/main.yml` with `delegate_to: localhost` on the few canasta_registry calls that need it.

## Background (from #426)

Both files had this pattern wrapping their canasta_registry calls:

```yaml
- name: Switch to local for registry queries
  ansible.builtin.set_fact:
    _saved_host: "{{ ansible_host | default('localhost') }}"
    _saved_conn: "{{ ansible_connection | default('local') }}"
    _saved_user: "{{ ansible_user | default('') }}"
    _saved_python: "{{ ansible_python_interpreter | default(...) }}"
    ansible_host: "localhost"
    ansible_connection: "local"
    ...
- name: Run the registry call
  canasta_registry: { ... }
- name: Restore connection
  ansible.builtin.set_fact:
    ansible_host: "{{ _saved_host }}"
    ...
```

Two flaws:

1. **Failure between save and restore leaves the play in a "switched" state.** Any task that fails (or aborts via `meta: end_play` etc.) between the save and restore strands the playbook with `ansible_connection=local` for downstream tasks that should have continued running against the original target.
2. **Save and restore drift.** The two lists must stay in sync (currently four fields each in two different blocks); a new field added to one but not the other silently breaks restoration. The audit caught one such drift (`_saved_python: ... default('auto')` vs `_ri_saved_python: ... default(ansible_playbook_python)`).

The same effect — running just the registry call against the controller, regardless of inbound connection — is already what most of the codebase does, with a single-task `delegate_to: localhost` plus a `vars` block. `roles/create/tasks/main.yml`'s "Check instance does not already exist" is the canonical example.

## Changes

- **`roles/common/tasks/resolve_instance.yml`**: drop the save/switch/restore. Still capture the inbound `ansible_host`/`ansible_connection` into `_ri_orig_*` because the resolution-path branching (remote vs localhost) needs to know which the operator targeted. Drop `_ri_saved_user` and `_ri_saved_python` (no longer needed since nothing's switching). Add `delegate_to: localhost` + `vars` to the three `canasta_registry` calls (`query_all`, `query_by_path`, `query`).
- **`roles/delete/tasks/main.yml`**: drop both dance instances (the "Handle already-removed instance" block at lines 226-247, and the standalone "Deregister instance from controller" block at lines 266-287). Each collapses to a single `canasta_registry` call with `delegate_to: localhost`.

Net: −29 lines, identical behavior.

## Test plan

- [x] `make validate` passes
- [x] `make test-unit` — 604 tests pass
- [ ] Manual: `canasta delete -i <id>` with the instance on a remote host (registry on controller, files on remote) still cleans up both registries.
- [ ] Manual: `canasta start/stop/restart -i <id>` (which depends on `resolve_instance`) for both local and remote-host instances still resolves correctly.

Closes #426.

### DIFF
--- a/roles/common/tasks/resolve_instance.yml
+++ b/roles/common/tasks/resolve_instance.yml
@@ -10,23 +10,17 @@
 # Input: id (optional)
 # Output: instance_id, instance_path, instance_orchestrator,
 #   instance_devmode, instance_managed_cluster
+#
+# Note: registry queries below all carry delegate_to: localhost so they
+# run against the controller's registry regardless of the inbound
+# connection state. We capture the inbound host/conn purely to drive the
+# resolution-path branching (remote vs localhost) — we do not switch
+# ansible_connection itself, which would have to be undone afterwards.
 
-# Save current connection state so we can restore after registry queries.
-# This is needed because resolve_instance may be called after
-# switch_connection.yml has set ansible_connection to ssh.
-- name: Save connection state before registry queries
+- name: Capture inbound connection state for resolution branching
   ansible.builtin.set_fact:
-    _ri_saved_host: "{{ ansible_host | default('localhost') }}"
-    _ri_saved_conn: "{{ ansible_connection | default('local') }}"
-    _ri_saved_user: "{{ ansible_user | default('') }}"
-    _ri_saved_python: "{{ ansible_python_interpreter | default(ansible_playbook_python) }}"
-
-- name: Switch to local for registry queries
-  ansible.builtin.set_fact:
-    ansible_host: "localhost"
-    ansible_connection: "local"
-    ansible_user: ""
-    ansible_python_interpreter: "{{ ansible_playbook_python }}"
+    _ri_orig_host: "{{ ansible_host | default('localhost') }}"
+    _ri_orig_conn: "{{ ansible_connection | default('local') }}"
 
 # Step 1: Determine which ID to look up
 - name: Use provided ID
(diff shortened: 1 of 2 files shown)


cicalese:
Verified empirically on a Linux VPS (`acknowledge.wiki`, Debian 13, native Docker) by running the remote-integration suite from PR #464 once against current `main` and once with this PR cherry-picked on top of `main` + #464.

**Result against current `main`** (#464's branch — save/switch/restore code):

```
Tests run: 10
Passed:    10
Failed:    0
```

**Result with this PR applied** (`delegate_to: localhost` + `vars:`):

```
Tests run: 10
Passed:    9
Failed:    1
```

The single failure is Test 8 (post-delete `canasta list`):

```
FAIL: list still shows deleted instance (output:
remote-test     testuser@canasta-test-remote  /tmp/canasta-test-data/remote-test
  COMPOSE       NOT FOUND)
```

That's the exact symptom b7d3ed2 fixed: `canasta delete -i <remote-id>` reports success but the controller's `conf.json` still has the entry, because the `canasta_registry: state=absent` task (now using `delegate_to: localhost` + `vars: { ansible_connection: local }`) was routed to the remote host instead of the controller. b7d3ed2's commit message:

> delegate_to: localhost was not reliably deregistering from the controller because set_fact on the play host contaminated the delegation target's connection vars.

Same symptom is reproducible today.

Test 6 (the new `config set` + restart scenario from #464 that I expected would catch the 0d392a6 multi-call bug) passed under both — so either that specific failure mode has been fixed by something else in the meantime, or my test doesn't exercise the exact path. The b7d3ed2 path is still live.

Recommendation: close this PR. If the dance is really bothersome, a follow-up could keep the connection switch but wrap it in `block`/`always:` for crash-safety and dict-saving for drift-safety — but the regression test in #464 should run green against any such refactor before it lands.

cicalese:
Superseded by #466, which addresses #426's underlying concerns (failure between save/restore, save/restore drift) by wrapping the pattern in `block`/`always:` with a single dict — without the regression this PR introduced. #466 scored 10/10 on the same regression suite this PR scored 9/10 on. Closing.

Issue #349, opened 2026-04-25
canasta install k8s should support --tls-san for remote-controller multi-node setups


cicalese:
### Problem

`canasta install k8s` (`roles/orchestrator/tasks/k8s_install_k3s.yml`) currently runs `curl -sfL https://get.k3s.io | sh -` with no extra args. The k3s API server's TLS cert ends up valid only for `127.0.0.1`, the node's internal IP, and the standard `kubernetes.*` SANs.

When the controller is a separate machine reaching the API via the cluster's public IP — i.e. a multi-node setup — `kubectl get nodes` from the controller fails with a cert-verification error.

`MULTI_NODE_TEST_PLAN.md` works around this by SSHing in and running the upstream installer with `--tls-san <NODE1_IP>` instead. `docs/multi-node.md` documents this same workaround and explicitly warns against using `canasta install k8s` for multi-node control-plane installs.

### Proposal

Add a `--tls-san <name>` (multi-valued) parameter to `canasta install k8s`, plumbed through to k3s as `INSTALL_K3S_EXEC="--tls-san <name>"` (or equivalent flag).

Optional refinement: when `--host` is set and the SSH connection target looks like an IP/hostname, default `--tls-san` to that target.

### Acceptance

- `canasta install k8s --host node1 --tls-san <PUBLIC_IP>` brings up k3s on a remote host with the public IP in the API server's TLS cert.
- `kubectl get nodes` from a separate controller succeeds without `-k` / `--insecure-skip-tls-verify`.
- `docs/multi-node.md`'s "Known gaps → `canasta install k8s` doesn't pass `--tls-san`" entry can be deleted.

PR #477, opened 2026-05-03
canasta scale: set web replica count for K8s instances (closes #392)

### LINKED ISSUE #392: canasta scale: no canasta-native way to set replica count; users have to bypass canasta with kubectl scale
## Background

Multi-node K8s with `--access-mode ReadWriteMany` exists specifically so you can run multiple replicas of the web pod across nodes (or other components — jobrunner, varnish — to a lesser extent). End-to-end testing of #349/#350 confirmed this works at the K8s layer:

```
$ kubectl scale -n canasta-testsite deploy/canasta-testsite-web --replicas=3
$ kubectl get pods -n canasta-testsite -l app.kubernetes.io/component=web -o wide
NAME                                    NODE
canasta-testsite-web-…-b6mqp            ip-172-31-1-11
canasta-testsite-web-…-cgjwz            ip-172-31-10-96
canasta-testsite-web-…-t8w82            ip-172-31-12-213
```

…but the only way to do this is direct kubectl. There is no canasta-native command.

## Gap

Looking at `meta/command_definitions.yml` and `canasta.py:SUBCOMMAND_GROUPS`:

- No `canasta scale` command
- No `--replicas` flag on `create`, `start`, `add`, or any other instance verb
- No mechanism to inspect or change replica counts through canasta at all

That means:

- Users on K8s have to bypass canasta to scale, which violates the "canasta is the source of truth for the instance" model.
- Any subsequent `canasta upgrade` re-renders the helm chart from per-instance values and resets replicas back to whatever the chart default is (1 today). The kubectl-scaled pod count silently disappears.
- The same gap exists on Compose (`docker-compose up --scale web=3` works there, but again outside canasta).

## Proposed shape

```
canasta scale --id <instance> [--component <name>] --replicas <N>
```

- `--component` defaults to `web` (the obvious primary scale target). Other valid values today: `jobrunner`, plus whatever Compose service names map to (`varnish` is a candidate but cache invalidation is more complex; `db`/`caddy` should probably reject as not-meaningful-to-scale).
- `--replicas N` sets the desired count.
- Persists to per-instance state so it survives `canasta upgrade`:
  - **K8s**: write `<component>.replicas: N` into the instance's helm values.yaml, then `helm upgrade --install`.
  - **Compose**: write the count into a per-instance file (or `.env`) that `canasta start` honors via `docker-compose up --scale`.
- Reading back: either a new `canasta scale --id X` (no `--replicas`) prints current counts, or extend `canasta list` to show them.

## Open questions

- Should `canasta scale` be allowed without `--access-mode ReadWriteMany` on K8s? With RWO, multiple web replicas would `Multi-Attach error` on the content PVCs. Probably reject up front when the storage class doesn't support RWM, with a usable error.
- Auto-scaling (HPA) is a separate, larger discussion — `canasta scale` is the manual-set primitive; HPA support is a follow-up.
- Compose interaction with the existing `start`/`stop` flow needs design — `start` would need to know the desired replica count.

## Why now

Without this, the multi-node K8s capability we just shipped is half-built: the platform supports multi-replica, but canasta users have no way to use it. They either edit values.yaml by hand or wear the `kubectl scale` footgun where the next `canasta upgrade` undoes their work.

cicalese:
## Summary

Closes the gap #392 reported. Multi-node K8s with `--access-mode ReadWriteMany` ships as of #349/#350, but the only way to actually set web replica count was direct `kubectl scale` — and any subsequent `canasta upgrade` silently undid it because the chart re-renders from `values.yaml` where the default is 1.

## Behavior

```
canasta scale -i <instance> --replicas <N>
```

- Reads the instance's `values.yaml`
- Sets `web.replicaCount: N`
- Runs `helm upgrade --install canasta-<id> <chart> -f values.yaml -f values-configdata.yaml --reset-values --wait --timeout 10m` — exact args helm_deploy.yml uses
- Result survives `canasta upgrade`

No-op fast path when the count already matches the stored value (skips helm entirely).

## Scope: Kubernetes only, by design

- **K8s only.** Not a Compose follow-up — Compose runs all replicas on a single host with no built-in cross-replica load balancing, so the spread-across-nodes / fault-tolerance / horizontal-capacity model that motivates this command doesn't apply. The error message on a Compose instance points operators at PHP-FPM worker limits as the right knob for "more concurrency" on Compose.
- **`--component web` only.** The chart currently parameterizes `replicas` only on `deployment-web.yaml`; jobrunner / caddy / varnish are hardcoded `replicas: 1`. Other components are rejected up front rather than silently editing values.yaml under a key the chart ignores. Adding more is a follow-up that needs both chart and CLI changes.
- **Auto-scaling (HPA) explicitly out of scope.** This is the manual-set primitive.

## Implementation

- New `direct_only` command in `direct_commands.py` (fits the pattern for K8s state ops that don't benefit from Ansible's per-task overhead).
- Two new helpers `_read_remote_or_local_file` / `_write_remote_or_local_file` so the command works against `-H <remote>` instances via SSH cat / cat-write.
- Validation: rejects non-K8s, rejects non-`web` components, requires integer `--replicas ≥ 1`.

## Test plan

- [x] `make test-unit` (668 passed, 8 new)
- [x] `make validate` (67 commands, 50 playbooks)
- [x] `make lint` (no new warnings)
- 8 unit tests cover: registration; orchestrator + component rejection; integer / range validation; no-op fast path; write+helm happy path (asserts on values.yaml content + helm command shape); and helm-failure-after-values-write (rc propagated, partial-state warning).
- [ ] Manual: smoke test on the existing k3s cluster on acknowledge.wiki — verify pods scale up/down and the count survives a `canasta restart`.

Closes #392

### DIFF
--- a/canasta.py
+++ b/canasta.py
@@ -802,7 +802,13 @@ def build_ansible_args(ansible_playbook, command_name, args, data):
         "ANSIBLE_SSH_ARGS",
         "-o StrictHostKeyChecking=accept-new "
         "-o UserKnownHostsFile=~/.ssh/known_hosts "
-        "-o ForwardAgent=yes",
+        "-o ForwardAgent=yes "
+        # Long-running remote commands (helm upgrade --wait, gitops
+        # init's git push, maintenance update) can outlast a NAT or
+        # firewall idle timeout. ServerAliveInterval keeps the SSH
+        # session warm so the parent doesn't see a "Broken pipe"
+        # while the remote is still working.
+        "-o ServerAliveInterval=30 -o ServerAliveCountMax=20",
     )
 
     # Hand the platform-correct config dir to Ansible. get_config_dir()
--- a/direct_commands.py
+++ b/direct_commands.py
@@ -2100,9 +2100,16 @@ def cmd_scale(args):
     inst_id, inst = _resolve_instance(args)
     orchestrator = inst.get("orchestrator", "compose")
     if orchestrator not in ("kubernetes", "k8s"):
+        # Kubernetes-only by design: Compose runs all replicas on the
+        # same host with no built-in cross-replica load balancing, so
+        # the spread-across-nodes / horizontal-capacity model that
+        # motivates this command doesn't apply. Bump PHP-FPM worker
+        # limits inside the image instead.
         print(
-            "Error: `canasta scale` is currently Kubernetes-only. "
-            "Compose support is tracked as a follow-up.",
+            "Error: `canasta scale` is Kubernetes-only by design. "
+            "On Compose, raise PHP-FPM worker limits inside the web "
+            "image rather than running multiple replicas of the same "
+            "container.",
             file=sys.stderr,
         )
         return 1
--- a/direct_commands.py
+++ b/direct_commands.py
@@ -2040,6 +2040,191 @@ def _kubectl_section(host, ns, cmd_args, label):
     return (label, out.rstrip())
 
 
+# ---------------------------------------------------------------------------
+# canasta scale
+# ---------------------------------------------------------------------------
+
+def _read_remote_or_local_file(path, host):
+    """Read a file from localhost or a remote host. Returns content or
+    None on error."""
+    if _is_localhost(host):
+        try:
+            with open(path) as f:
+                return f.read()
+        except OSError as e:
+            print("Error reading %s: %s" % (path, e), file=sys.stderr)
+            return None
+    rc, content = _ssh_run(host, "cat %s 2>/dev/null" % _shell_quote(path))
+    if rc != 0:
+        print("Error reading remote %s" % path, file=sys.stderr)
+        return None
+    return content
+
+
+def _write_remote_or_local_file(path, host, content):
+    """Write content to path on localhost or a remote host. Returns True
+    on success."""
+    if _is_localhost(host):
+        try:
+            with open(path, "w") as f:
+                f.write(content)
+            return True
+        except OSError as e:
+            print("Error writing %s: %s" % (path, e), file=sys.stderr)
+            return False
+    target = _resolve_ssh_target(host)
+    ssh_cmd = ["ssh"] + _ssh_args() + [target, "cat > %s" % _shell_quote(path)]
+    try:
+        result = subprocess.run(
+            ssh_cmd, input=content, text=True,
+            capture_output=True, timeout=30,
+        )
+    except (subprocess.TimeoutExpired, OSError) as e:
+        print("Error writing remote %s: %s" % (path, e), file=sys.stderr)
+        return False
+    if result.returncode != 0:
+        print(
+            "Error writing remote %s: %s"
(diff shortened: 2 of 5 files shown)


cicalese:
Smoke-tested end-to-end on the existing single-node k3s cluster on acknowledge.wiki (laptop → remote K8s cp via SSH, ssh-agent loaded with the operator's key).

**Setup:** `canasta create -H acknowledge.wiki -i scaletest -w main -n acknowledge.wiki --orchestrator kubernetes` → instance up with default `web.replicaCount: 1`.

**Test sequence:**

1. **Downscale 3 → 1** (cluster had 3 from earlier test runs):
   ```
   $ canasta scale -i scaletest --replicas 1
   Release "canasta-scaletest" has been upgraded. Happy Helming!
   …
   STATUS: deployed
   REVISION: 4
   Scaling web to 1 replica(s)…
   Scaled web to 1 replica(s). Persisted to /home/cicalese/canasta/scaletest/values.yaml.
   ```
   Two old pods → `Terminating`, one → `Running`. `values.yaml` shows `replicaCount: 1`.

2. **Upscale 1 → 2:**
   ```
   $ canasta scale -i scaletest --replicas 2
   …
   REVISION: 5
   Scaled web to 2 replica(s). Persisted to /home/cicalese/canasta/scaletest/values.yaml.
   ```
   New pod joined (52s old next to the existing 4h pod).

3. **No-op fast path** (scale to current count):
   ```
   $ canasta scale -i scaletest --replicas 2
   web.replicaCount is already 2; nothing to do.
   ```
   No helm round-trip — confirmed by lack of helm output.

4. **Unsupported component:**
   ```
   $ canasta scale -i scaletest --component varnish --replicas 2
   Error: only 'web' supports replica scaling. Component 'varnish' is not supported.
   ```

**Two real bugs surfaced and fixed in 887965d:**

a. `_ssh_run`'s 30-second subprocess timeout was firing well before helm's own `--wait --timeout 10m` completed. Command falsely reported "helm upgrade failed" while helm was still working. Fix: `cmd_scale` now runs the helm command via `subprocess.call` directly (no Python-side timeout) and streams output to the operator.

b. The remote SSH session itself dropped mid-helm-upgrade with "Connection closed by remote host. client_loop: send disconnect: Broken pipe" — a NAT / firewall idle TCP drop. Fix: append `-o ServerAliveInterval=30 -o ServerAliveCountMax=20` to both the `ANSIBLE_SSH_ARGS` default in `canasta.py` (playbook path) and the fallback in `direct_commands._ssh_args` (direct-command path). 30-second keepalive probes for up to 10 minutes before SSH gives up — matches helm's `--timeout 10m`.

`test_default_ansible_ssh_args_carries_required_options` extended to assert `ServerAliveInterval` is present so silently losing it doesn't resurrect the failure mode.

Cleanup: instance deleted; cluster infrastructure (k3s + Argo CD) left up for future K8s tests.

Issue #443, opened 2026-04-29
API edits don't invalidate Varnish cache (browser edits do)


cicalese:
## Summary

Edits made via the MediaWiki API (`action=edit`) do **not** invalidate the Varnish cache, while edits via the browser UI **do**. The result is stale content after any automation-driven edit (bots, gitops-style sync, scripted updates, `wiki_publish.py`-equivalent tooling), persisting until either a manual PURGE/ban hits the right cache key or Varnish is restarted.

## Reproduction (observed in canasta.wiki this session)

1. Edit a page via the API as a logged-in admin (no `bot=1` flag). For example, via Python:
   ```python
   r = post({
       "action": "edit",
       "title": "Help:Common tasks",
       "text": new_text,
       "token": csrf,
       "summary": "...",
       "assert": "user",
       "format": "json",
   })
   # Returns Success, new revid in response.
   ```
2. After several minutes, fetch the page over HTTPS:
   ```
   curl -sI 'https://example.com/wiki/Help:Common_tasks' | grep -iE 'age|x-varnish'
   age: 14461
   x-varnish: 145078 143381
   ```
   Two `x-varnish` values → cache HIT. `age:` shows the cached object is hours old (predates the API edit). New content not visible to the public.
3. Make a trivial edit through the browser UI on the same page (add and remove a space, Save).
4. Re-fetch:
   ```
   age: <small>
   x-varnish: <new-id>
   ```
   Cache invalidated; new content visible immediately.

## What we ruled out

- **JobRunner backed up.** `canasta maintenance script showJobs.php` returns `0` — no pending jobs.
- **Varnish ACL stale.** Manual PURGE from inside the `web` container (where the `acl purge { "web"; }` rule allows it) returns `200 Purged`, proving the ACL accepts traffic from the right source.
- **Wrong wiki / wrong host.** The wiki is in this instance's `wikis.yaml`, manual PURGE targeting the right `Host: <wiki-host>` and path returns 200.

## What's still unclear

- Whether `MultiHttpClient`'s PURGE fanout from MediaWiki is firing at all for API edits, vs firing but failing silently (it discards errors).
- Whether the deferred-update / post-send pipeline for API requests differs from the browser request lifecycle in modern MediaWiki (≥ 1.39) in a way that drops `HtmlCacheUpdater::purgeUrls()` on the API side.
- Whether this is upstream MediaWiki behavior or a Canasta-specific config interaction.

## Wiring (for reference)

[...]

PR #549, opened 2026-05-07
varnish: dedupe default.vcl via symlink (single source of truth)

cicalese:
## Summary

Replace `roles/orchestrator/files/helm/canasta/files/default.vcl` (a byte-for-byte duplicate of `instance_template/config/default.vcl`) with a symlink to that canonical. Single source of truth; future fixes touch one file.

## Why

Surfaced as a follow-up to #443 / #548. Both VCL files had to be edited identically for the fix to land. With the symlink, drift is structurally impossible — there's only one file.

The `instance_template` direction is canonical because it's the operator-facing template (the file Compose users see at `~/canasta/<id>/config/default.vcl` after `canasta create`). The K8s Helm chart's `files/` dir is internal plumbing that operators don't normally touch.

## Verified consumers follow symlinks

- '''Helm''' — `helm template roles/orchestrator/files/helm/canasta/` renders the ConfigMap with the full canonical VCL content (including #443's RFC 1918 ACL block).
- '''Ansible `copy`''' — the chart-deploy task copies the chart directory to `{{ instance_path }}/_chart` with default `local_follow: yes`. Symlinks resolve at copy time; the destination gets a real `default.vcl` file. Deployed chart shape unchanged.

## Test plan

- [x] `make test-unit` — 779 passed.
- [x] `make validate` — clean.
- [x] `helm template` — renders the canonical VCL through the symlink.
- [ ] One real `canasta create -o k8s` to confirm the chart-copy + ConfigMap reach the cluster as before. (Heavily exercised today on EC2 hosts; the chart-copy code path didn't change, only the VCL source location did.)

## Out of scope

- `my.cnf` is also duplicated (1 line: `[client]`) — not worth the symlink overhead at that size.
- `Caddyfile.global`, `Caddyfile.site`, settings files, `.env` — none of those are duplicated; they live in `instance_template/` only and flow to K8s via different mechanisms (configMap rendering, K8s Secrets) without a second canonical copy.


cicalese:
Superseded by a cleaner architectural fix: drop the K8s-side default.vcl entirely and have Ansible's k8s_sync_config.yml push the per-instance VCL into configData.varnish, mirroring how Caddyfile is already handled. No symlink, no duplicate, and K8s operators gain per-instance VCL customization (parity with Compose). Will open a new PR shortly.

PR #570, opened 2026-05-23
Add wiki-check command to check if wikis are accessible

### LINKED ISSUE #569: Create "canasta wiki-check" command
There is already a "canasta docker" command that checks the health of the Canasta installation. However, what is still needed is a command to check the wiki itself, to make sure that MediaWiki is running correctly, as well as that wiki-related processes like backups are being done. Ideally, a "canasta wiki-check" command would check most or all of the following:
- That a page exists at _WikiURL_/wiki/Main_Page (or _WikiURL_/Main_Page, if short URLs are being used) - or whatever the main page is called
- That the API is accessible, and can handle simple actions like "action=query&list=allpages"
- Pages are cached correctly
- Search works
- Thumbnail generation works
- File and DB backup works, if it has been enabled
- GitOps backup works, if it has been enabled
- Emailing works, if SMTP has been enabled
- Actions that cause actual changes work, such as:
   - Editing/creating pages
   - Deleting pages
   - File uploads
 
For actions that cause changes, it may make sense to have a separate flag, so calling "canasta wiki-checker" by itself does not have any side effects.

naresh-kumar-babu:
Partial Fix #569.
Check that the wiki base URL is reachable.


cicalese:
Thanks for the contribution. I have several questions/suggestions.

## Standalone or add to an existing command?

Existing commands:

| command | what it checks |
|---|---|
| `list` | container/pod **running** (`docker compose ps -q web` / k8s `readyReplicas`) |
| `status` | orchestrator/infra state: pods, PVCs, ingress/domain, TLS cert |
| `doctor` | host tooling (Docker/Helm/git…) |

None of them make an HTTP request to MediaWiki, so a container can read as "RUNNING" while the wiki itself returns 500 or Caddy isn't routing. This PR's HTTP-200 check fills that gap. It overlaps conceptually with `canasta status`, which is where a user looks for "is my instance up?". I think it would make sense to add this part of #569 and possibly other non-destructive checks to `canasta status`.

## Don't hard-code `Main_Page`

The main page title is configurable (`$wgMainPage` / `MediaWiki:Mainpage`, and localized — e.g. "Main Page" with a space, or a non-English title), so `/wiki/Main_Page` and `/Main_Page` will false-negative on any wiki whose main page has been renamed. Name-agnostic alternatives:

- **Hit the wiki root** (`WikiURL/`): MediaWiki redirects `/` to whatever the main page is, so a 200 after redirects proves reachability without knowing the title.
- **Use the API**: `…/api.php?action=query&meta=siteinfo&format=json` — a stable endpoint independent of the main page name, and you can assert the response actually looks like MediaWiki (`batchcomplete`/`query`) rather than just "something returned 200". This is what the integration suite's `wait_for_wiki` already uses, and #569 lists an API check explicitly.

Switching to root-redirect and/or `api.php` also simplifies `_build_urls`.

## Other notes

- **Dead code:** `port` is computed in `_build_urls` but never used — the port is already part of `netloc`/`host`.
- **localhost vs. domain:** when the instance is on localhost but the `wikis.yaml` URL is a real domain, the `urllib` path resolves that domain via public DNS rather than the local instance. `wait_for_wiki` handles this by requesting `127.0.0.1:<port>` with a `Host:` header — worth mirroring for the localhost path.
- **Cert verification asymmetry:** the localhost path uses `ssl._create_unverified_context()` (skips verification), but the remote path runs `curl` without `-k`, so a staging/self-signed cert (`CANASTA_STAGING_CERTS`) produces a false negative on remote hosts. Pick one policy.
- **Tests:** `_build_urls` is a pure function and trivially unit-testable — a good addition (especially after the rework off `Main_Page`).

cicalese:
With respect to where this wiki accessibility check belongs, I think it is fine to be in the wiki-check command. But, it would be good if it were also written so it could be easily called from `canasta list` as well. `canasta list` currently lists which instances are known, whether each is running, and what wikis are registered for each instance. It would be helpful if it also reported whether each wiki is accessible.

direct_commands/wiki_check.py
```diff
@@ -0,0 +1,100 @@
...
+
+        port = None
+        if ":" in parsed.netloc:
+            port_value = parsed.netloc.rsplit(":", 1)[-1]
+            if port_value.isdigit():
+                port = port_value
+
+        base_urls = ["%s://%s" % (protocol, host) for protocol in ["https", "http"]]
+
+    wiki_main_page_suffixes = ["/wiki/Main_Page", "/Main_Page"]
```
jeffw16: Consts should not be inline in functions; they should be in a clear section either in a consts file (preferred) or near the top of the file in a consts section

jeffw16 (changes requested):

direct_commands/wiki_check.py
naresh-kumar-babu: Updated

cicalese:
Thanks for the rework — this addresses most of the round-1 feedback well.

Two things I'd still like to see before this is ready:

**1. Probe `api.php` instead of the root.** Hitting `…/api.php?action=query&meta=siteinfo&format=json` avoids issues from the redirect, and you can assert the body actually looks like MediaWiki (`batchcomplete`/`query`) rather than just "something returned 200." This is what the integration suite's `wait_for_wiki` already does, and it also simplifies `_build_urls` since you no longer depend on root-redirect behavior.

**2. Minor:** a bare `localhost` URL with no port in `wikis.yaml` takes the direct-`urlopen` branch and hits port 80, ignoring `HTTP_PORT`/`HTTPS_PORT` — a false negative if the instance runs on a non-default port. Worth routing through the same env-port logic.

The `list` integration we discussed can be a follow-up — having `_check_url(wiki_url, host, instance_path)` as a reusable seam is enough for now.

naresh-kumar-babu:
> Thanks for the rework — this addresses most of the round-1 feedback well.
> 
> Two things I'd still like to see before this is ready:
> 
> **1. Probe `api.php` instead of the root.** Hitting `…/api.php?action=query&meta=siteinfo&format=json` avoids issues from the redirect, and you can assert the body actually looks like MediaWiki (`batchcomplete`/`query`) rather than just "something returned 200." This is what the integration suite's `wait_for_wiki` already does, and it also simplifies `_build_urls` since you no longer depend on root-redirect behavior.
> 
> **2. Minor:** a bare `localhost` URL with no port in `wikis.yaml` takes the direct-`urlopen` branch and hits port 80, ignoring `HTTP_PORT`/`HTTPS_PORT` — a false negative if the instance runs on a non-default port. Worth routing through the same env-port logic.
> 
> The `list` integration we discussed can be a follow-up — having `_check_url(wiki_url, host, instance_path)` as a reusable seam is enough for now.

Updated.

cicalese:
Thank you for your contribution. I have some reservations about proliferation in the number of commands, but we'll see if this command grows additional useful functionality.

PR #582, opened 2026-06-03
gitops: resilient join checkout + re-attach already-registered host on --reinit

### LINKED ISSUE #663: gitops --reinit can destroy the local .git on failure
### Summary

`gitops --reinit` is meant to recover an instance after a previous `init`/`join` died mid-run. But the reinit cleanup deletes the instance's `.git` **before** the operation succeeds, so a failed `--reinit` — the exact case the flag exists for — can leave the instance worse off than before.

### Details

`_reinit_cleanup.yml` removes `.git` (among other gitops files) early in the flow:

- **`gitops join --reinit`** — cleanup runs *before* the `git clone` (the clone targets a temp dir). If the clone fails (bad repo URL, auth, network), the instance's old `.git` is already gone, leaving it with **no `.git` at all**.
- **`gitops init --reinit`** — cleanup runs *before* `git init` / commit / push. If a later step fails (e.g. the push is rejected), the old `.git` — which may hold **local gitops commits that were never pushed** — has already been discarded.

In both cases the destructive step happens before the operation is known to succeed.

### Proposed fix

Make `--reinit` non-destructive:

- **join**: perform the clone *before* the destructive cleanup, so a failed clone leaves the existing `.git` untouched.
- **init**: back up `.git` before cleanup and restore it if the re-init fails.

Apply to both the Compose and Kubernetes paths (`join.yml`, `join_kubernetes.yml`, `init_compose.yml`, `init_kubernetes.yml`).

Surfaced while working on #571 / #582 (gitops join robustness).

### LINKED ISSUE #668: gitops: internal git commits fail when user has commit.gpgsign=true but gpg is unavailable
## Problem

When the environment running canasta has `commit.gpgsign = true` in its git config (commonly a global setting) but `gpg` is not installed or not configured, canasta's own internal `git commit` operations fail:

```console
$ canasta gitops join --reinit -n wiki-staging.nichework.com --repo mail.nichework.com:wiki.nichework.com --key ~/.config/canasta/my-key --id nichework
Error: The command exited with a non-zero return code. (cmd: git commit -m Add host wiki-staging.nichework.com)
error: cannot run gpg: No such file or directory
error: gpg failed to sign the data:
(no gpg output)
fatal: failed to write commit object
```

Canasta's internal commits are bookkeeping for its own gitops repo and should never depend on the user's commit-signing configuration. A user who signs their personal commits should not have canasta operations break as a side effect.

## Scope

All internal `git commit` invocations under `roles/gitops/`:

- `roles/gitops/tasks/join.yml`
- `roles/gitops/tasks/join_kubernetes.yml`
- `roles/gitops/tasks/init_compose.yml`
- `roles/gitops/tasks/init_kubernetes.yml`
- `roles/gitops/tasks/push_compose.yml`
- `roles/gitops/tasks/push_kubernetes.yml`
- `roles/gitops/tasks/fix_submodules.yml`

## Suggested fix

Pass `-c commit.gpgsign=false` to canasta's internal `git commit` calls so signing config never affects them. (Consider `-c tag.gpgsign=false` too if any tags are created.)

## Reproduction

Set `git config --global commit.gpgsign true` in an environment without a working gpg, then run any `canasta gitops` operation that commits (`init`, `join`, `push`).

Surfaced during validation of #582 / #571.

cicalese:
Addresses #571, closes #663, closes #668.

This PR carries three related fixes hardening `gitops join`/`--reinit`, surfaced in sequence while the reporter tested the branch.

### Problem

`gitops join` moves the cloned `.git` into the instance directory and runs `git checkout HEAD -- .`, which rewrites every tracked path. That unlinks files under `config/` (`Caddyfile.global`, `Caddyfile.site`, `composer.local.json`) even when their content already matches HEAD. Those files are bind-mounted and can be owned by the container user, so the account running the join may be unable to unlink them — aborting with an opaque `non-zero return code (cmd: git checkout HEAD -- .)`.

### Change

- Check out `.gitattributes` first so git-crypt decrypts `hosts/**` on the subsequent checkouts.
- Diff against HEAD and check out only the files that actually differ, so unchanged files are never unlinked (resolves the common all-defaults case with no permission error).
- On a permission failure, name the blocked files and give the remediation (`stop` / `chown` / `--reinit`) instead of failing with a bare git error.
- Applied to both the Compose (`join.yml`) and Kubernetes (`join_kubernetes.yml`) join paths.

### Problem

With the checkout fixed, the reporter got past it and hit the next wall: `Host '…' already exists in the gitops repository`. Once a host is registered (a prior join died after the push, or the instance's local gitops state was lost), there was no way to reconnect it — `--reinit` wiped only **local** state, while the registration in the **remote** repo kept tripping the duplicate-host guard. `--reinit`'s whole purpose is mid-run recovery, so it shouldn't dead-end on exactly that case.

### Change

- Under `--reinit`, the duplicate-host guard no longer fails — an already-registered host is treated as a **re-attach**.
- "Add host to hosts.yaml" now **upserts** (drops any existing same-name entry before appending), so re-attach replaces the entry in place instead of duplicating it.
- A clean re-attach can reproduce the repo tree unchanged, leaving nothing staged; the commit step is now gated on staged changes so the join doesn't abort on git's "nothing to commit". (Push stays a harmless no-op in that case.)
- `--reinit`'s help text documents the re-attach behavior.
- Applied to both the Compose and Kubernetes join paths.

### Problem

`--reinit` exists to recover an instance after a previous init/join died mid-run, but the reinit cleanup deleted `.git` **before** the operation was known to succeed:

- **join `--reinit`** deleted `.git` before the `git clone` (which targets a temp dir); a failed clone left the instance with **no `.git` at all**.
- **init `--reinit`** deleted `.git` before re-creating the repo and pushing; a failed push **discarded local gitops commits that were never pushed**.

So a failed `--reinit` — the exact case the flag exists for — could leave the instance worse off than before.

### Change

- **join**: run `_reinit_cleanup` *after* the clone succeeds, and gate the ".git already exists" guard on `not reinit`, so a failed clone leaves the existing `.git` untouched.
- **init**: move `.git` aside before re-init and restore it from a `block`/`rescue` if the re-init fails, then re-raise. Restore preserves the repo and its history; the working tree can be rebuilt with `git checkout -- .`.
- Applied to both the Compose and Kubernetes paths.

### Problem

While testing `join --reinit`, the reporter hit a commit failure on a host with `commit.gpgsign = true` in its git config but no usable `gpg`:

```
error: cannot run gpg: No such file or directory
error: gpg failed to sign the data
fatal: failed to write commit object
```

Canasta's bookkeeping commits to its **own** gitops repo were inheriting the user's signing configuration, so a personal `commit.gpgsign=true` (with gpg unavailable) aborted `init`/`join`/`push` entirely.

### Change

- Run every internal `git commit` with `-c commit.gpgsign=false`, so canasta's own commits never depend on the user's signing config. Applied to all call sites across both orchestrators (`init`, `join`, `push`, `fix-submodules`).
- Adds `test_gitops_no_gpgsign` (unit) — recursively scans every `roles/gitops` task and fails if a `git commit` call site omits the flag, so a new one can't reintroduce the dependency.

## Tests

Extends `test_gitops_join` (integration) — stands up a second instance, seeds it with a stale copy of a tracked settings file, joins it to the first instance's gitops repo (a **local bare repo**, no network), and asserts a repo-only file is materialized, the diverging tracked file is replaced, the other host's vars survive, and **re-attaching with `--reinit` upserts the host rather than duplicating or rejecting it**.

Adds `test_gitops_reinit_preserve_git` (unit) — structural guards that join runs `_reinit_cleanup` after the clone, the ".git exists" guard is gated on `not reinit` (both orchestrators), and init backs up `.git` before dispatch and restores it from the rescue.

`make lint`, `make validate`, and `make test-unit` (994 passed) are clean, and `gitops-join` passes locally.

## Status

The reporter confirmed Fix 1 resolves the original permission error; Fix 2 addresses the follow-up duplicate-host failure they then hit. Fix 3 closes the related data-safety gap (#663). Fix 4 closes the gpg-signing failure (#668) surfaced while testing `--reinit`. Still seeking validation on a real two-host Linux setup where the container-owned `config/` condition from #571 reproduces (the permission-failure branch can't be exercised on the single-host CI/dev harness); the reporter was running a stale local build for the re-attach attempt, so the Fix 2 re-attach path also still needs a clean rebuild to confirm.


hexmode:
``` console
$ git branch
* fix/gitops-join-resilient-checkout
  main

$ docker build -t ghcr.io/canastawiki/canasta-ansible:latest .
STEP 1/15: FROM python:3.12-slim
...
STEP 15/15: ENTRYPOINT ["/opt/canasta-ansible/canasta-native"]
--> Using cache 1850cb65562e95a7afce6f6a79e86fd6dec21e4f2b6e854f6b2efa2e38a09044
COMMIT ghcr.io/canastawiki/canasta-ansible:latest
--> 1850cb65562e
Successfully tagged ghcr.io/canastawiki/canasta-ansible:latest
1850cb65562e95a7afce6f6a79e86fd6dec21e4f2b6e854f6b2efa2e38a09044

$ sudo cp canasta-docker /usr/local/bin/canasta 

$ pushd
~/client/nichework/wiki.nichework.com ~/work/code/Canasta-CLI

$ canasta gitops join -n web.nichework.com --repo="$(git remote get-url origin)" --key ~/.config/canasta/my-nichework-key -i nichework
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
Error: Required parameter '--repo' is missing

**Comment by hexmode** [comment]
ping me if I did something wrong.

**Comment by cicalese** [comment]
Nothing wrong with the build — the join just never ran. Those two `fatal:` lines are coming from your own `$(git remote get-url origin)`, not from canasta.

You ran it from inside the instance directory (`wiki.nichework.com`), which isn't a git checkout, so `git remote get-url origin` failed, printed to stderr, and the command substitution expanded to an empty string. So canasta received `--repo=""` and correctly rejected it as missing.

`--repo` wants the **shared gitops config repo URL** — the same one you passed to `gitops init` for the first host (e.g. `git@github.com:org/config.git`). The instance directory isn't a clone of that repo, so deriving it from `origin` there was never going to work.

Just pass it explicitly:

```console
$ canasta gitops join -n web.nichework.com \
    --repo git@github.com:youorg/nichework-config.git \
    --key ~/.config/canasta/my-nichework-key -i nichework
```

That should finally exercise the resilient-checkout path on your container-owned `config/`.

Side note: an empty `--repo` slipping through is messy - the error is accurate but the `fatal:` lines above it are misleading. I'll improve the error message.

hexmode:
yep, you were right.  Found another issue.  Create a new issue?

```console
tmp$ canasta gitops join -n wiki.nichework.com --repo mail.nichework.com:wiki.nichework.com --key ~/.config/canasta/my-nichework-key 
Error: No Canasta instance is registered for this directory (/home/mah/client/nichework/tmp).
Specify --id <id>, or cd into an instance directory.
Registered instances: nichework

tmp$ canasta gitops join -n wiki.nichework.com --repo mail.nichework.com:wiki.nichework.com --key ~/.config/canasta/my-nichework-key --id nichework
Error: GitOps already initialized (/home/mah/client/nichework/wiki.nichework.com/.git exists). This instance is already connected to a gitops repository. Pass --reinit to wipe the local gitops state and re-clone from the remote (typical when a previous init or join died mid-run).

tmp$ canasta gitops join --reinit -n wiki.nichework.com --repo mail.nichework.com:wiki.nichework.com --key ~/.config/canasta/my-nichework-key --id nichework
Error: Host 'wiki.nichework.com' already exists in the gitops repository. Choose a different name or remove the existing host first.

tmp$ canasta gitops join --reinit -n wiki-staging.nichework.com --repo mail.nichework.com:wiki.nichework.com --key ~/.config/canasta/my-nichework-key --id nichework
Error: The command exited with a non-zero return code. (cmd: git commit -m Add host wiki-staging.nichework.com)
error: cannot run gpg: No such file or directory
error: gpg failed to sign the data:
(no gpg output)
fatal: failed to write commit object

**Comment by cicalese** [comment]
Good progress — but most of that sequence is a **stale build**, not a new bug. The tell is this error:

> Host 'wiki.nichework.com' already exists in the gitops repository. Choose a different name or remove the existing host first.

That's the message from *before* Fix 2. The current branch tip says:

> Choose a different name, remove the existing host first, **or pass `--reinit` to re-attach this host.**

Here's the gotcha: CI only builds/pushes the docker image on `main`, version tags, or manual dispatch — **never for PR branches**. So the only way PR code reaches the `canasta-docker` wrapper is a local `docker build` from the checked-out branch, and your earlier build came from a checkout that predated Fix 2 (`a20dcb1`) and Fix 3 (`c981bc0`, both 2026-06-13). You got Fix 1 — that's why your original permission error cleared — but not the re-attach fix, so step 3 (`--reinit -n wiki.nichework.com`, same name) failed instead of re-attaching.

Two ways forward:

**Easiest for PR testing — use the native wrapper** (runs the checkout directly, no image rebuild between code changes):

```console
$ git fetch origin && git reset --hard origin/fix/gitops-join-resilient-checkout
$ python3 -m venv .venv && .venv/bin/pip install -r requirements.txt   # one-time
$ ./canasta-native doctor          # confirms host deps
$ ./canasta-native gitops join --reinit -n wiki.nichework.com … --id nichework
```

**Or stick with the docker wrapper**, but rebuild after updating:

```console
$ git fetch origin && git reset --hard origin/fix/gitops-join-resilient-checkout
$ docker build --no-cache -t ghcr.io/canastawiki/canasta-ansible:latest .
```

Either way, please retry the **original same-name** form (`--reinit -n wiki.nichework.com`) — that's the re-attach path I most need validated.

The gpg failure at the end *is* a real, separate issue, though — you've got `commit.gpgsign = true` but no usable gpg, so canasta's internal commits can't sign. Our own bookkeeping commits shouldn't depend on your signing config; filed as #668 and we'll disable signing on them. (You can unblock yourself meanwhile with `git config --global commit.gpgsign false` or by installing gpg.)

Thanks for testing and reporting these issues!

PR #771, opened 2026-06-19
fix: maintenance exec --stdin-file (direct-exec path: stdin + quoting)

### LINKED ISSUE #770: maintenance exec --stdin-file has no effect and mangles quoted args (direct-exec path)
### Summary

`canasta maintenance exec` with a command does not run through Ansible —
`handle_interactive_exec` (canasta.py) execs `docker compose exec` /
`kubectl exec` / `ssh` directly via `os.execvp`. Two bugs in that path:

1. **`--stdin-file` is ignored.** The flag (added in #768) was wired only
   into the Ansible exec path (`roles/orchestrator/tasks/exec.yml`), which
   `maintenance exec <cmd>` never reaches. The direct path also runs
   `docker compose exec` without `-T`, so even an inherited pipe wouldn't
   survive. Net: the executed command gets empty stdin.

2. **Quoted arguments are mangled.** The direct path splits a `--`
   passthrough command with a naive `str.split()`, shredding quoted tokens
   on whitespace. A `--summary` with spaces or a page title with spaces
   (e.g. `'Template:User receipts'`) breaks apart.

### Impact

Piping content to a maintenance script is broken. Concretely,
`maintenance exec --stdin-file page.txt -- php maintenance/run.php edit
--user=WikiSysop --summary='Initial import' 'Main Page'` writes nothing
useful: empty stdin (bug 1), and the content lands on a page literally
titled `'--summary=Initial` (bug 2).

### Found by

A real deploy that imports starter wiki pages via `run.php edit` (reading
content from stdin). Pages imported empty and `cargoRecreateData` reported
the tables "not declared in any template" because the templates had no
content.

### Fix

Honor `stdin_file` in `handle_interactive_exec` (add `-T` for compose /
`-i` for k8s / `ssh -T`, and dup the file onto fd 0 before exec), and use
`shlex.split` for the passthrough command so quoted args survive.

cicalese:
Closes #770.

`canasta maintenance exec <cmd>` runs through `handle_interactive_exec`,
which execs docker/kubectl/ssh directly via `os.execvp` — it never reaches
the Ansible exec path where #768 wired `--stdin-file`. Two fixes there:

## 1. --stdin-file now takes effect (`e57fe7a`)
- Compose: add `-T` (no TTY) so a pipe survives; `dup2` the file onto fd 0.
- K8s: `kubectl exec -i` (drop the TTY) + `dup2`.
- Remote: `ssh -T` + `dup2`; ssh forwards stdin to `docker compose exec -T`.
- Interactive behavior (no `--stdin-file`) is unchanged (`-it`/`-t`).

## 2. Quoted args survive (`a29f225`)
The direct path split the `--` passthrough with a naive `str.split()`,
shredding quoted tokens on whitespace. Use `shlex.split` so a spaced
`--summary` or a page title with spaces stays one argv element.

## Why both, found together
A live deploy importing starter pages via `run.php edit` (content on stdin):
with bug 1 the pages imported empty; once stdin flowed, bug 2 wrote them to
a page titled `'--summary=Initial`. Fixed, the import lands correctly and
`cargoRecreateData` builds the tables.

## Tests
+4 regression tests over `handle_interactive_exec` (stdin `-T`/`-i` flags,
the stdin redirect, and quote preservation). `make validate`, `make lint`,
and the full unit suite (1163 tests) pass.

## Note
These two bugs both stem from `maintenance exec` having a second,
Ansible-bypassing code path that re-implements the orchestrator dispatch.
Consolidating the direct and Ansible exec paths (keeping `os.execvp` only
for genuinely interactive sessions) would prevent this class of drift — a
worthwhile follow-up, out of scope here.

### DIFF
--- a/canasta.py
+++ b/canasta.py
@@ -744,12 +744,17 @@ def handle_interactive_exec(args):
       - -s given, no command -> interactive /bin/bash in that service
       - -s given, with command -> exec in that service
     """
+    import shlex
     service = getattr(args, "service", None) or ""
     exec_args = getattr(args, "exec_args", None) or []
     if isinstance(exec_args, list):
         command = exec_args
     else:
-        command = exec_args.split() if exec_args else []
+        # exec_args arrives as a string when main() collected it from a `--`
+        # passthrough. Split it like a shell would so quoted arguments (a
+        # --summary with spaces, a page title with spaces) survive as single
+        # argv elements — a naive .split() would shred them on whitespace.
+        command = shlex.split(exec_args) if exec_args else []
 
     # No service flag AND no command -> list services (let Ansible handle it)
     if not service and not command:
--- a/canasta.py
+++ b/canasta.py
@@ -713,6 +713,28 @@ def resolve_instance(instance_id=None):
     sys.exit(1)
 
 
+def _redirect_stdin_from_file(path):
+    """Point fd 0 at `path` so the about-to-be-exec'd command reads it as
+    stdin. Used by `maintenance exec --stdin-file`. No-op when path is unset.
+
+    This runs in the direct-exec path (handle_interactive_exec os.execvp's
+    docker/kubectl/ssh, replacing this process), so there is no shell to do
+    a `< file` redirect — we dup the file onto stdin ourselves.
+    """
+    if not path:
+        return
+    try:
+        fd = os.open(path, os.O_RDONLY)
+    except OSError as exc:
+        print(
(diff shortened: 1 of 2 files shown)


cicalese:
The duplication that caused both of these bugs is now tracked as #773 (consolidating the direct-exec and Ansible exec paths, plus the same pattern in pod resolution / registry / validation). This PR is the targeted bug fix; #773 is the follow-up refactor.

PR #785, opened 2026-06-19
fix: deliver custom extensions/skins to the web pod on K8s (#784)

### LINKED ISSUE #784: k8s: custom extensions/skins are never delivered to the container (user-extensions/user-skins PVCs are never populated)
### Severity: high — custom extensions silently do not work on Kubernetes

Adding a custom extension to an instance's `extensions/` directory works on
Compose but **silently does nothing on Kubernetes**. The extension code never
reaches the web container, so `wfLoadExtension(...)` fatals (or the extension is
simply absent).

### Root cause

- **Compose:** `roles/orchestrator/files/compose/docker-compose.yml` bind-mounts
  `./extensions` → `/var/www/mediawiki/w/user-extensions`. The CanastaBase
  entrypoint (`create-symlinks.sh`) then symlinks `user-extensions/*` into the
  MediaWiki `extensions/` dir, so MediaWiki loads them.
- **k8s:** `deployment-web.yaml` mounts `user-extensions` (and `user-skins`)
  from **PVCs** (`<fullname>-extensions` / `-skins`). The web Deployment's only
  initContainers are `assemble-config` (config → /config) and `wait-for-db` —
  **nothing populates the extensions/skins PVCs** from the instance's
  `extensions/`/`skins/` directories.

`k8s_sync_config.yml` syncs only config (wikis.yaml/.env/settings/\*) →
ConfigMaps. The generic `copy_into_container.yml` (kubectl cp) exists but is
called only by `import_database.yml`. The gitops repo tracks `extensions/`, but
no task carries that content into the PVC. So the PVC stays empty,
`create-symlinks.sh` has nothing to symlink, and the extension is absent.

### Reproduce

1. `canasta create -o kubernetes -i x ...`
2. Copy a custom extension into `<instance>/extensions/MyExt/` and load it in a
   settings fragment (`wfLoadExtension('MyExt')`).
3. `canasta restart -i x`.
4. The web pod CrashLoopBackOffs on the missing extension;
   `kubectl exec <web> -- ls /var/www/mediawiki/w/user-extensions` is empty.

(Found while deploying a real multi-component app to a 2-node k3s cluster.)

### Fix direction

Add a k8s step that populates the `user-extensions` / `user-skins` PVCs from the
instance's `extensions/` / `skins/` dirs (a no-op on Compose, where the
bind-mount already does this). The PVCs are mounted in the web pod, so the
content can be delivered via `kubectl cp` into the running pod (before the
rollout that loads the extensions) or via a short-lived Job/pod that mounts the
PVCs. This should run on `create` and on `restart`/config-apply (whenever
`extensions/`/`skins/` may have changed) — mirroring how the bind-mount keeps
them live on Compose.

cicalese:
## Problem

Closes #784.

On Docker Compose the `./extensions` and `./skins` bind-mounts make custom extensions and skins live in the container. On Kubernetes those paths are PVCs (`user-extensions` / `user-skins`) that **nothing populated from the instance directory**. An extension dropped into `extensions/` therefore never reached the web pod, and CanastaBase's `create-symlinks.sh` had nothing to link into `extensions/`. If a settings fragment then `wfLoadExtension()`'d that extension, the web pod CrashLoopBackOff'd.

## Fix

Add `roles/orchestrator/tasks/k8s_sync_user_dirs.yml`: before the Helm rollout, a short-lived helper pod mounts the `canasta-<id>-extensions` and `canasta-<id>-skins` PVCs and `kubectl cp`s each top-level entry from the instance's `extensions/` and `skins/` dirs into them.

- Runs **before** `helm_deploy`, so the new web pod starts with the code already present — no CrashLoopBackOff when settings load a custom extension.
- The helper pod mounts the PVCs **directly**, so the sync is independent of the web pod's lifecycle and recovers even from an already-broken (crash-looping) instance.
- Wired into the **restart** flow (`start.yml`) — the path every add-extension workflow takes. (Create doesn't need it: the PVCs don't exist until Helm creates them, and `extensions/` is empty on a fresh instance; the real flow is always "add extensions → restart".)
- Additive copy (no prune in v1).

## Testing

- `make validate` (77 commands), `make lint` (yamllint + ruff), `make test-unit` (1195 tests) — all pass.
- **Live single-node k3s:** moved an installed instance's extensions aside inside the PVC (broken state), ran `canasta restart`, and confirmed the sync task restored `user-extensions/<ext>`, CanastaBase symlinked `extensions/<ext> -> ../user-extensions/<ext>`, and the web pod returned to `1/1 Running` — with no manual workaround.

### DIFF
--- a/roles/orchestrator/tasks/k8s_sync_user_dirs.yml
+++ b/roles/orchestrator/tasks/k8s_sync_user_dirs.yml
@@ -1,0 +1,125 @@
+---
+# Populate the user-extensions / user-skins PVCs from the instance's
+# extensions/ and skins/ directories (Kubernetes).
+#
+# On Compose the ./extensions and ./skins bind-mounts make custom
+# extensions/skins live in the container. On K8s those paths are PVCs
+# that nothing else populates from the instance dir, so custom
+# extensions/skins would never reach the web pod (the entrypoint's
+# create-symlinks.sh then has nothing to symlink into extensions/). #784
+#
+# Runs BEFORE the Helm rollout so the new web pod starts with the code
+# already in the PVC — avoids a CrashLoopBackOff when a settings fragment
+# wfLoadExtension()s a not-yet-present extension. A short-lived helper pod
+# mounts the PVCs, so this is independent of the web pod's lifecycle (works
+# on first create when no web pod exists yet, and on restart).
+#
+# Input: instance_path, instance_id (or id), _k8s_namespace (optional)
+
+- name: Resolve namespace + PVC names for user-dir sync
+  ansible.builtin.set_fact:
+    _ud_namespace: "{{ _k8s_namespace | default('canasta-' ~ (instance_id | default(id))) }}"
+    _ud_release: "canasta-{{ instance_id | default(id) }}"
+
+# extensions -> <release>-extensions PVC @ /sync/extensions
+# skins      -> <release>-skins PVC      @ /sync/skins
+- name: Scan instance user dirs for content
+  ansible.builtin.find:
+    paths: "{{ instance_path }}/{{ item }}"
+    file_type: any
+    depth: 1
+    excludes: ['.gitignore', '.gitkeep', '.git']
+    hidden: false
+  register: _ud_scan
+  loop: ['extensions', 'skins']
+
+- name: Build the list of dirs that actually have content
+  ansible.builtin.set_fact:
+    _ud_dirs: >-
+      {{ _ud_scan.results
+         | selectattr('matched', 'defined')
+         | selectattr('matched', 'gt', 0)
+         | map(attribute='item') | list }}
+
+- name: Sync user dirs into their PVCs (only when there is content)
+  when: _ud_dirs | length > 0
+  block:
+    - name: Render user-dir sync helper pod
+      ansible.builtin.copy:
+        dest: "{{ instance_path }}/.user-dir-sync-pod.yaml"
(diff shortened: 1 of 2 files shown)


cicalese:
### Behavior note: on K8s, adding an extension/skin requires a `canasta restart` (unlike Compose)

There is a real orchestrator difference here, worth documenting:

- **Compose — no restart needed.** The `./extensions` bind-mount writes straight into the container's `user-extensions`, and `monitor-directories.sh` (`inotifywait -m` on `user-extensions`/`user-skins`) auto-creates the `extensions/<ext>` symlink the moment the dir appears (and reverts to `canasta-extensions`, or removes the symlink, on delete). `Settings.php`'s `wfLoadExtension()` is then live on the next request.
- **K8s — restart required (this PR).** Nothing populates the `user-extensions` PVC from the instance dir, so there is nothing for the in-pod watcher to react to. This PR's sync writes the code into the PVC from a separate helper pod *before* the Helm rollout, and the fresh pod links it via `create-symlinks.sh` at entrypoint. (The running pod's `inotify` can't be relied on to observe a write made by another pod into a shared/NFS-backed PVC, so rolling the pod is the reliable path.)

So the rule differs by orchestrator: **Compose picks up a new extension automatically; K8s needs a `canasta restart`.** A possible future enhancement would be to make the K8s path restart-free (e.g. have the sync target the live pod), but that's out of scope for this fix.

PR #795, opened 2026-06-20
feat(create): add preflight disk space check for container storage

### LINKED ISSUE #793: 01 Storage pre-flight check and documentation
## Summary
Canasta currently has no pre-flight check for available container storage space.
On systems where container storage (`~/.local/share/containers/` for rootless
Podman, `/var/lib/docker/` for Docker) resides on a small root partition,
pulling Canasta's ~2.3 GB image can exhaust disk space before any meaningful
data is added, making the instance unstable.

## Environment
- Debian 12 Bookworm, r7g.large EC2 (50 GB root, +100 GB data volume)
- Podman 4.3.1 (rootless) via podman-docker shim
- Canasta CLI v4.2.0

## Observed
- Root disk reached 100% usage (47 GB / 49 GB) after pulling images and
  running one container for a few minutes.
- Container storage occupied 43 GB on the root partition.
- No warning from `canasta create` about available space before pulling images.

## Root Cause
Canasta does not check available storage before pulling images or creating
containers. The `docker system df` (or `podman system df`) command is available
to query this, but Canasta never calls it.

## Workaround Applied
We pre-configure Podman's storage location to a separate 100 GB data volume
via `~/.config/containers/storage.conf` before the first `podman pull`. This is
handled by our infrastructure provisioning script (user-data.sh).

## Suggested Fix
1. Add a pre-flight check early in `canasta create` that runs
   `docker system df` (or equivalent for the active runtime) and warns if the
   backing filesystem for container storage has less than e.g. 20 GB free.
2. Document the storage layout concern in the setup guide — container images
   are large (~2.3 GB for Canasta) and the default storage location may exhaust
   small root partitions on VPS/EC2 instances.

This is a feature request / documentation enhancement, not a bug fix.

hexmode:
## Summary

Adds a pre-flight check during `canasta create` that warns when Docker storage has less than 20 GB free.

## Problem

Canasta images are ~2.3 GB. On instances with small root partitions (e.g. 50 GB EC2 root volumes), pulling images and running containers can exhaust disk space before any meaningful data is added. There is currently no warning from `canasta create` about available storage space.

## Changes

Three new tasks in `roles/create/tasks/main.yml` (Compose-only, after the memory check):

1. **Get Docker root directory** — queries `docker info` for the Docker Root Dir path
2. **Check available space** — runs `df` on that path to get free space in 1K blocks
3. **Warn if below 20 GB** — emits a warning message with the path and available space

The check is best-effort: failures are ignored silently so it works across Docker CE, podman-docker, and rootless setups. It only runs for Compose orchestrator (K8s has its own storage considerations).

## Related

Closes #793.

### DIFF
--- a/roles/orchestrator/tasks/create_preflight.yml
+++ b/roles/orchestrator/tasks/create_preflight.yml
@@ -192,14 +192,16 @@
     # operator can free space or move container storage to a larger
     # volume without re-provisioning the host.
     #
-    # The Docker Root Dir is parsed from the `docker info` output already
-    # captured by the availability check above, so no extra docker call
-    # is needed.
-    - name: Resolve Docker root directory
+    # The storage root is parsed from the `docker info` output already
+    # captured by the availability check above, so no extra runtime call
+    # is needed. Docker CE reports it as `Docker Root Dir`; the
+    # podman-docker shim (rootless Podman) reports `graphRoot` instead,
+    # so match either.
+    - name: Resolve container storage root
       ansible.builtin.set_fact:
         _docker_root_dir: >-
           {{ (_docker_check.stdout | default('')
-             | regex_findall('Docker Root Dir:\s*(.+)') | first | default('')) | trim }}
+             | regex_findall('(?:Docker Root Dir|graphRoot):\s*(.+)') | first | default('')) | trim }}
 
     # `df --output=avail` is GNU coreutils only (BusyBox/BSD lack it);
     # `failed_when: false` keeps the check best-effort so it silently
--- a/roles/orchestrator/tasks/create_preflight.yml
+++ b/roles/orchestrator/tasks/create_preflight.yml
@@ -184,3 +184,39 @@
     - name: Record host-gateway fallback IP
       ansible.builtin.set_fact:
         _host_gateway_fallback: "{{ _host_gateway_probe.stdout | default('') | trim }}"
+
+    # --- Compose-only storage preflight ---
+    # Canasta images are ~2.3 GB; with database volumes, logs, and build
+    # cache, container storage can exhaust a small root partition. Warn
+    # early so the operator can configure storage on a larger volume.
+    - name: Get Docker root directory
+      ansible.builtin.shell:
+        cmd: "set -o pipefail && docker info 2>/dev/null | awk -F': ' '/Docker Root Dir/{print $2}'"
+      register: _docker_root_dir
+      changed_when: false
+      failed_when: false
+
+    - name: Check available disk space on Docker storage
+      ansible.builtin.shell:
+        cmd: >-
+          set -o pipefail && df --output=avail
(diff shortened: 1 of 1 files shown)


cicalese:
Thanks for this, @hexmode — useful check. I rebased the branch onto current `main` and made a few adjustments so it could land without extra work on your end. Your original commit is preserved (authored to you); my follow-up is a separate commit. Summary of what changed:

- **Rebased / relocated.** `main.yml` had moved its pre-flight checks into `roles/create/tasks/_preflight.yml` since this branch was cut (the `_mem_check` anchor your diff referenced no longer lives in `main.yml`), so the branch no longer applied cleanly. The storage check now sits in `_preflight.yml` alongside the existing Compose pre-flight probes (memory, ports 80/443, k3s).
- **Reuse `docker info`.** `_preflight.yml` already runs `docker info` once (the Docker-available gate, `_docker_check`). The Docker Root Dir is now parsed from that captured output via `regex_findall`, dropping the two extra `docker info` invocations.
- **GiB labeling.** Threshold and display are binary (1024-based), so the message now says GiB to match the math.
- **Comments.** Noted that `df --output` is GNU-coreutils-only (best-effort `failed_when: false` no-ops on BusyBox/BSD), and why this warns rather than hard-fails like the memory check (disk is recoverable mid-flight).

`make lint` and `make test-unit` (1215 tests) pass. Could you give it a review when you have a moment? Happy to adjust anything.

cicalese:
This implements the pre-flight check half of #793.

cicalese:
Added one more commit (`a68e330`): the check now also matches `graphRoot` in `docker info`, not just `Docker Root Dir`. The original #793 report was from rootless Podman via the podman-docker shim, where `docker info` is `podman info` and reports the storage path as `graphRoot` — so without this the check would have silently no-op'd on exactly the environment that prompted the issue. Verified the regex against both `Docker Root Dir: …` (Docker CE) and `graphRoot: …` (Podman) output, plus the no-match case.

cicalese:
@hexmode this is rebased onto current `main` and ready for a look. Both halves of #793 are now covered:

- **Pre-flight check** — in this PR, in `roles/create/tasks/_preflight.yml` alongside the other Compose pre-flight probes. Matches both Docker CE (`Docker Root Dir`) and rootless Podman (`graphRoot`).
- **Docs** — added a "Disk space and relocating container storage" subsection to [Help:Storage](https://canasta.wiki/wiki/Help:Storage) (Docker `data-root` / Podman `graphroot`, plus the 20 GiB warning).

Could you test on your end? The case that matters most is your original environment — **rootless Podman via the podman-docker shim on a small root partition**. I verified the path-parsing regex against sample `docker info`/`podman info` output, but I don't have a podman host to exercise the full `canasta create` flow, so a live run there would confirm the warning actually fires (and reports the right `graphRoot` path and free-space number). A normal Docker CE run on a <20 GiB-free disk to confirm the wording would be great too.

`make lint` and `make test-unit` (1215) pass locally. Let me know if anything needs adjusting.

PR #796, opened 2026-06-20
docs: add Cloudflare TLS termination pattern guide

hexmode:
## Summary

Documents the configuration pattern for running Canasta behind Cloudflare TLS termination.

## Problem

Canasta defaults to `CADDY_AUTO_HTTPS=on`, which causes Caddy to attempt Let's Encrypt certificates and emit HTTP-to-HTTPS redirects. When Cloudflare terminates TLS at the edge, this creates double-redirect loops and unnecessary certificate attempts.

## Solution

No code changes needed — `CADDY_AUTO_HTTPS=off` already works correctly. This PR adds a documentation guide covering:

- The problem (redirect loops, failed ACME challenges)
- Two-step solution: `CADDY_AUTO_HTTPS=off` + optional Caddyfile.site headers
- Security considerations (restrict port 80 to Cloudflare IPs, SSL/TLS mode guidance)
- Verification commands to confirm Caddy is not redirecting

## Related

Refs Cloudflare TLS termination documentation enhancement.


cicalese:
Thanks for writing this up, Mark — the Cloudflare guidance is genuinely valuable and something we keep getting bitten by, so I wanted it captured well.

A couple of structural things led me to move it rather than merge it here:

- **`docs/` is an auto-generated output directory.** It only holds the CLI command reference produced by `make docs` from `meta/command_definitions.yml` (git-ignored except `.gitkeep`). A hand-written guide there isn't discoverable by users and drifts out of sync with the docs wiki. I'm adding a `docs/README.md` + CONTRIBUTING note so this is clearer going forward (the old CONTRIBUTING line saying "design docs live in `docs/`" was misleading — that's on me).
- **All narrative docs live on canasta.wiki.** Cloudflare is common enough to deserve its own page, so I've created **[Help:Using Canasta with Cloudflare](https://canasta.wiki/wiki/Help:Using_Canasta_with_Cloudflare)** as a hub that gathers the previously-scattered advice, and linked it from Networking and TLS, Troubleshooting, and the user-guide index.

The page keeps your core content (the `CADDY_AUTO_HTTPS=off` pattern, the redirect-loop diagram, origin-IP lockdown, verification curls) with one correction worth flagging: the SSL/TLS mode determines the Canasta setting. `CADDY_AUTO_HTTPS=off` is correct for Cloudflare **Flexible** mode, but **Full / Full (strict)** instead wants Caddy to keep its own (default) Let's Encrypt certificate — running Full with auto-HTTPS off is exactly what produces the 525s. I also added the Bot-Fight-Mode-blocks-ACME 525 gotcha we've hit in the field.

Could you take a look at the new page and let me know if anything's missing? I've closed this PR in favor of the wiki page — please re-open it if you think anything's been lost in the move.

PR #802, opened 2026-06-21
fix(create): support podman < 4.4 for host-gateway extra_hosts

### LINKED ISSUE #797: docker-compose.yml template uses host-gateway, incompatible with podman < 4.4
## Description

The docker-compose.yml template generated by `canasta create` includes:
```yaml
extra_hosts:
  - "gateway.docker.internal:host-gateway"
```

And `/opt/canasta-ansible/canasta-docker` uses:
```bash
MOUNTS+=(--add-host=host.docker.internal:host-gateway)
```

`host-gateway` was added in podman 4.4.0, but Debian 12 (bookworm) ships podman 4.3.1. This causes the compose step to fail with:

```
Error: invalid IP address in add-host: "host-gateway"
```

## Environment
- Canasta CLI v4.2.0 (native, commit b6b6bd1)
- Canasta image: 3.5.10
- OS: Debian 12 (arm64)
- Podman: 4.3.1+ds1-8+deb12u1+b3
- Orchestrator: compose

## Workaround

We patched the templates after install:
```bash
GATEWAY_IP=$(podman network inspect podman --format "{{range .Subnets}}{{.Gateway}}{{end}}")
find /opt/canasta-ansible -type f \( -name "*.yml" -o -name "canasta-docker" \) \
  -exec sed -i "s/:host-gateway/:$GATEWAY_IP/g" {} \;
sed -i "s/host-gateway/$GATEWAY_IP/g" /opt/canasta-ansible/canasta-docker
```

## Suggested Fix

Either:
1. Use `${DOCKER_GATEWAY:-host-gateway}` with a fallback to the actual gateway IP
2. Detect podman version and use the appropriate value
3. Document the podman 4.4+ requirement

cicalese:
## Summary

`canasta create` on Compose fails on podman < 4.4 (e.g. Debian 12's 4.3.1) with:

```
Error: invalid IP address in add-host: "host-gateway"
```

podman only learned the `host-gateway` magic value in 4.4, but the generated `docker-compose.yml` and the `canasta-docker` kubeconfig path both hard-code it.

## Fix

Keep the portable `host-gateway` token everywhere it's supported (Docker, podman ≥ 4.4) and only substitute a resolved literal IP on podman < 4.4:

- **`docker-compose.yml`** — the web service's `extra_hosts` now uses `${HOST_GATEWAY:-host-gateway}`. Unset means the magic value (current behavior).
- **`_preflight.yml`** — on Compose targets, probe the podman version; when it's < 4.4, resolve the bridge gateway IP (`podman network inspect podman --format '{{range .Subnets}}{{.Gateway}}{{end}}'`) into a fact. Empty for Docker and podman ≥ 4.4.
- **`_env_update.yml`** — pin `HOST_GATEWAY` in `.env` only when that fact is non-empty (a data condition, so it's a no-op on Docker / modern podman).
- **`canasta-docker`** — the same podman < 4.4 fallback for the `--add-host` in the dockerized-CLI kubeconfig path.

The probe is best-effort (`failed_when: false`): a Docker-only host, a missing `podman` binary, or an unexpected network layout all leave `HOST_GATEWAY` unset and fall back to today's behavior.

## Testing

- `make lint`, `make validate`, and the full unit suite (1215 tests) pass.
- Version-comparison logic verified across 4.3.1 / 4.3.0 / 3.4.4 / 4.4.0 / 4.4.1 / 4.5.0 / 5.0.0.
- The Jinja `{% raw %}` block was confirmed to preserve the Go-template braces in the rendered shell command.

Not yet exercised live on a podman < 4.4 host; the resolution command mirrors the workaround proven in the issue.

Closes #797

### DIFF
--- a/canasta-docker
+++ b/canasta-docker
@@ -229,7 +229,21 @@ with open(dst, 'w') as f:
 PYEOF
         fi
         if [[ -f "$CANASTA_KUBE_CONFIG" ]]; then
-            MOUNTS+=(--add-host=host.docker.internal:host-gateway)
+            # "host-gateway" is understood by Docker and podman >= 4.4 but
+            # rejected by podman < 4.4; fall back to the bridge gateway IP
+            # there so `--add-host` doesn't abort the run.
+            host_gw="host-gateway"
+            if command -v podman &>/dev/null; then
+                pv="$(podman --version 2>/dev/null | awk '{print $3}')"
+                if [[ -n "$pv" \
+                    && "$(printf '%s\n4.4.0\n' "$pv" | sort -V | head -n1)" == "$pv" \
+                    && "$pv" != "4.4.0" ]]; then
+                    gw="$(podman network inspect podman \
+                        --format '{{range .Subnets}}{{.Gateway}}{{end}}' 2>/dev/null)"
+                    [[ -n "$gw" ]] && host_gw="$gw"
+                fi
+            fi
+            MOUNTS+=(--add-host=host.docker.internal:"$host_gw")
         fi
     fi
     if [[ -f "$CANASTA_KUBE_CONFIG" ]]; then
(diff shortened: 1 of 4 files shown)


cicalese:
@hexmode Please test

PR #960, opened 2026-07-01
fix(upgrade): make MySQL 8-to-MariaDB migration work (port from Go CLI)

### LINKED ISSUE #959: MySQL to MariaDB upgrade fails
Since .rocksdb doesn't exist (it never does in a normal MySQL 8 datadir — that file only appears with the MyRocks storage engine, which Canasta
  doesn't use), the check fell through to echo mariadb. So _is_mysql8 was set to false, and the whole migration block — dump, clear volume,
  reimport — was skipped silently. No error, no failure. It just decided your MySQL 8 data "wasn't MySQL 8" and moved on.

  That's the root cause. Your data directory is the direct proof: it's unmistakably MySQL 8 (mysql.ibd, #innodb_redo, undo_001/002), yet the
  detector saw no .rocksdb and concluded "mariadb."

  What then broke MariaDB (and where the two aria files come from)

  With the migration skipped, the upgrade continued normally:

  1. canasta upgrade force-refreshed the managed docker-compose.yml (roles/upgrade/tasks/main.yml:97). The new managed compose hardcodes image: 
  docker.io/library/mariadb:11.4. So your db service's image was swapped from mysql:8.0 → mariadb:11.4 right there.
  2. The upgrade restarted the stack. A fresh MariaDB 11.4 container came up pointed at the same volume — full of MySQL 8 files.
  3. MariaDB's entrypoint saw a non-empty datadir → your log: MariaDB upgrade information missing, assuming required → it launched a temporary 
  server to run mariadb-upgrade.
  4. Booting that temporary server, MariaDB's very first act is to initialize its Aria engine (which backs its own system tables) — that's when
  it wrote aria_log.00000001 and aria_log_control (your two files, timestamped 16:46, after the 16:32 crash log). Then it hit MySQL 8's InnoDB
  tablespaces, couldn't read that format → Unable to start server.

  So those two aria_* files are the fingerprint of MariaDB's failed boot, not of the migration. They're the last thing MariaDB managed to do
  before choking on the InnoDB files — which is also why your actual data (the 16:13 InnoDB files) is untouched and recoverable.

  In one line

  The migration checks for the wrong marker file (.rocksdb), so it never recognized your MySQL 8 data; the upgrade then swapped the image to
  MariaDB and restarted, and MariaDB crashed trying to boot on data it can't read — leaving those two stray Aria log files behind.

cicalese:
Fixes #959.

## Problem

The MySQL 8.0-to-MariaDB upgrade migration left the database unbootable after `canasta upgrade`:

```
[Note] [Entrypoint]: MariaDB upgrade information missing, assuming required
[ERROR] [Entrypoint]: Unable to start server.
```

Root cause: the Ansible port of this migration regressed the retired Go CLI's working implementation (`Canasta-Go`'s `cmd/upgrade/mysql_to_mariadb.go`) in **three** independent ways:

1. **Detection** keyed off `/var/lib/mysql/.rocksdb` — a MyRocks marker absent from every stock MySQL 8 datadir — so the migration was *always* silently skipped. The upgrade then force-refreshed the compose to `mariadb:11.4` and restarted, and MariaDB crashed booting on the un-migrated MySQL 8 datadir.
2. **Dump mechanism**: it dumped from whatever container the instance compose started and restarted via that same (still MySQL 8) compose, so even with the marker fixed it couldn't produce a working MariaDB and couldn't recover an instance already swapped to MariaDB.
3. **Dump scope**: `mysqldump --all-databases` dragged MySQL 8's system schema (`mysql`/`sys`/`*_schema`, whose grant-table layout differs) into MariaDB, breaking the import.

(The detection bug dates to the migration's original commit, but stayed dormant until #615 made the compose refresh authoritative on upgrade — see the [backstory](https://github.com/CanastaWiki/Canasta-CLI/pull/960#issuecomment-4858853555).)

## Fix — port the Go behavior faithfully

- Detect via a read-only `mysql.ibd` check against the data volume from a throwaway container, so it works whether the DB is up, stopped, or crash-looping (and can recover an instance already swapped to MariaDB).
- Dump from a temporary `mysql:8.0` container the migration starts itself, independent of the instance compose.
- Dump only user databases (exclude `mysql`/`sys`/`*_schema`); MariaDB rebuilds its own system tables on the fresh datadir. Carry stored programs with `--routines`/`--triggers`/`--events`.
- Force-refresh the compose to the MariaDB image between clearing the volume and the restart. Preserve the host-side dump on failure for manual recovery.

## Tests

Neither the Go CLI nor the Ansible CLI had any test for this migration — that's why it shipped broken.

- **Unit** (`tests/unit/test_mysql_to_mariadb_migration.py`): structural guards for the `mysql.ibd` marker, the dedicated `mysql:8.0` dump container, user-DB-only dump, stop→dump and clear→refresh→start ordering, and dump-preserved-on-failure.
- **Integration** (`upgrade-mysql-to-mariadb`): seeds a genuine MySQL 8.0 datadir with a sentinel row, runs `canasta upgrade`, and asserts the row survives in MariaDB — the end-to-end coverage that was missing. It fails on the old code (migration skipped → MariaDB crash-loops → the query errors).

Full unit suite passes (1496). The integration test requires Docker + a `mysql:8.0` pull and is push-to-main-gated, so it hasn't run in CI yet; a live e2e against real MySQL 8 data is still recommended before relying on it in a release.

### DIFF
--- a/roles/upgrade/tasks/migrations/mysql_to_mariadb.yml
+++ b/roles/upgrade/tasks/migrations/mysql_to_mariadb.yml
@@ -1,47 +1,58 @@
 ---
-# Migration: Detect MySQL 8.0 data and migrate to MariaDB.
+# Migration: Detect MySQL 8.0 data and migrate it to MariaDB.
+# Ported from the retired Go CLI (cmd/upgrade/mysql_to_mariadb.go).
 # Compose-only; the orchestrator guard is owned by the caller
 # (roles/orchestrator/tasks/upgrade_db_migration.yml), so no
 # instance_orchestrator switch is needed here.
 # Input: instance_path
+#
+# Approach: detection is a read-only file check against the data volume, so
+# it works whether the DB is running, stopped, or crash-looping (e.g. a
+# MariaDB container already failing to boot on MySQL 8 data). The dump runs
+# against a throwaway mysql:8.0 container we start ourselves, so it does not
+# depend on the instance's compose still pinning MySQL 8 — and it can recover
+# an instance whose compose was already swapped to MariaDB. Only user
+# databases are dumped; MySQL 8's system schema (mysql/sys/*_schema) is left
+# behind, since MariaDB rebuilds its own grant tables on a fresh datadir.
 
-- name: Check for MySQL 8.0 data
+- name: MySQL 8.0 to MariaDB migration
   block:
-    # Detection (and the migration itself) requires a live db container to
-    # exec into. A stopped instance has no db service, so skip rather than
-    # abort the upgrade. ignore_errors on the include_tasks below would not
-    # help: it does not propagate to the shell task inside exec.yml, so the
-    # running-state is guarded explicitly here instead.
-    - name: Default MySQL 8 detection flag
+    # Compose v2 derives the project name by lowercasing the instance
+    # directory name and stripping characters outside [a-z0-9_-], then
+    # prefixes the volume. Match that exactly or the volume is not found.
+    - name: Resolve the MySQL data volume name
       ansible.builtin.set_fact:
-        _is_mysql8: false
+        _mig_volume: >-
+          {{ (instance_path | basename | lower
+              | regex_replace('[^a-z0-9_-]', '')) }}_mysql-data-volume
 
-    - name: Check whether the db service is running
+    # mysql.ibd is MySQL 8.0's system-schema InnoDB tablespace; MariaDB never
+    # creates it (it keeps the mysql schema in Aria tables under mysql/). Its
+    # presence is the definitive MySQL 8 marker. A throwaway container reads
+    # the volume without needing the DB service up.
+    - name: Detect MySQL 8.0 data on the volume
       ansible.builtin.command:
-        cmd: "docker compose ps -q db"
-        chdir: "{{ instance_path }}"
-      register: _db_ps
+        cmd: >-
+          docker run --rm -v {{ _mig_volume }}:/data
+          alpine test -f /data/mysql.ibd
+      register: _mig_detect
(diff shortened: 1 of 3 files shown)


cicalese:
## Backstory: why this only started failing recently

The detection bug is old but was dormant until a second change activated it.

- **2026-03-28 (`1b49247`)** — the migration shipped with the wrong marker (`.rocksdb`, a MyRocks file absent from stock MySQL 8 datadirs), so it was *always* skipped. Harmless at the time: on upgrade, `write_stack_files` copied the managed `docker-compose.yml` with `force: false` (no-clobber), so a legacy MySQL 8 instance kept its own `mysql:8.0` compose. The DB image was never swapped, the wiki kept running, and nobody noticed the migration was a no-op.

- **2026-06-10 (#615, `b134e33`)** — upgrade began **force-refreshing** the managed compose (`stack_files_force: true`) and recreating on a compose change. That overwrites the instance's `docker-compose.yml` with the managed one, which hardcodes `mariadb:11.4`. Now the DB image gets swapped out from under the unconverted MySQL 8 datadir, so MariaDB boots on data it can't read and dies (`MariaDB upgrade information missing` → `Unable to start server`).

So #615 didn't create the detection bug — it's three months older and latent — but making the compose refresh authoritative on upgrade is what turned the silently-skipped migration into a boot crash. Before #615: skipped migration, no harm. After #615: skipped migration, dead DB.

This is also why the fix has an ordering half: since #615 made the compose swap authoritative on upgrade, the migration must now control image state itself (dump on MySQL 8 → refresh compose → restart on MariaDB), rather than assuming the compose stays put.

Issue #1021, opened 2026-07-04
Add canasta image push / image prune for the in-cluster registry


cicalese:
## Motivation

The in-cluster image registry (fixed ClusterIP 10.43.0.2:5000, deployed at k8s-cp install) is currently reachable only through two internal side-channels: \`canasta create --build-from\` (create-time web image) and \`build:\` sidecars (pushed on every start). There is no user-facing way to get a locally built image into the cluster, even though the CLI's own docs tell operators to do exactly that: \`canasta rebuild\` says "Kubernetes instances must build and push their custom image to a registry manually."

Concrete cases blocked today:
- A derived elasticsearch image (e.g. the ICU analysis plugin for an alternate collation) on a Kubernetes instance: \`CANASTA_ELASTICSEARCH_IMAGE\` accepts any pullable ref, but a locally built tag isn't pullable by the cluster, so the operator needs an external registry account to move an image into their own cluster.
- Updating a custom web image on a running k8s instance outside of \`canasta upgrade\` (the standalone equivalent of \`canasta rebuild\`, which is Compose-only).
- Iterating on a custom Canasta/CanastaBase build against a live k8s instance without recreating it through \`--build-from\`.

Separately, nothing ever garbage-collects the registry: every re-push of a \`:local\` tag orphans the previous manifest's layers on the 10Gi PVC, which therefore only grows.

## Proposal

A new \`image\` subcommand group with two verbs, Kubernetes-only (on Compose, locally built tags are directly usable and the commands say so):

\`canasta image push <context> --name <name>[:tag]\` — ensure the registry exists (reuse \`k8s_ensure_registry\`), \`docker build\` the context tagged \`localhost:5000/<name>:<tag>\` (default tag \`local\`), push via the control plane's loopback hostPort, and print the cluster-pullable ref \`10.43.0.2:5000/<name>:<tag>\` for use with \`canasta config set CANASTA_IMAGE=…\` / \`CANASTA_ELASTICSEARCH_IMAGE=…\` or a sidecar \`image:\`.

\`canasta image prune\` — run the registry's garbage collector with \`--delete-untagged\` inside the registry pod, then restart the deployment (clears the registry's in-memory blob cache so subsequent pushes are consistent). Reclaims the layers orphaned by repeated same-tag pushes.

[...]

PR #1030, opened 2026-07-04
k8s: add a startup probe so elasticsearch survives slow starts

### LINKED ISSUE #1026: k8s elasticsearch crash-loops on modest hosts: liveness probe allows ~90s startup, no startupProbe
Found in a hands-on k8s e2e (single-node cluster, 8-core/16GB Debian 13 host).

With `CANASTA_ENABLE_ELASTICSEARCH=true` on Kubernetes, the elasticsearch pod never becomes Ready: the kubelet kills it on a loop (`Killing ... failed liveness probe`, exit code 143, 7 restarts and counting).

The chart's ES container has:
- Liveness probe `http /_cluster/health, initialDelaySeconds=0, period=30s, failureThreshold=3` → the container must serve HTTP within ~90s of start or be killed
- `resources.limits.cpu: 500m` and `ES_JAVA_OPTS=-Xms256m -Xmx256m`

ES 7.17 startup under a 500m CPU limit takes longer than 90s on this host, so every restart hits the same deadline — a permanent crash loop. The image itself is fine: the identical image run unconstrained via `docker run` on the same host serves `/_cat/plugins` in ~15s.

Fix suggestion: add a `startupProbe` (e.g. same endpoint, `periodSeconds: 10`, `failureThreshold: 30` → 5 minutes of grace) so the liveness probe only takes over after startup, which is the standard pattern for slow-starting JVM containers. Possibly also revisit the 500m CPU limit for the startup phase.

Noticed while validating #1021's `image push` with a derived ES image (`CANASTA_ELASTICSEARCH_IMAGE` pointed at the in-cluster registry): the pod pulled the pushed image by digest correctly — the delivery path works — but stock ES would crash-loop identically under these probe settings.

cicalese:
Closes #1026.

On Kubernetes, the elasticsearch container had only a liveness probe (`/_cluster/health`, period 30s, failureThreshold 3, no initial delay), giving it ~90 seconds from container start to serve HTTP. Under the chart's 500m CPU limit, ES 7.17 routinely takes longer than that on modest hosts, so the kubelet killed it on a permanent loop (exit 143, `failed liveness probe, will be restarted`) — the pod never became Ready. The image itself is fine: the identical image served `/_cat/plugins` in ~15s when run unconstrained on the same host.

Fix: a `startupProbe` (same endpoint, period 10s, failureThreshold 30) owns the boot window — up to 5 minutes — and the liveness probe only takes over once it passes, the standard pattern for slow-starting JVM containers. The readiness probe's `initialDelaySeconds` is dropped since readiness doesn't run until the startup probe succeeds.

Found in the hands-on k8s e2e of #1021 while validating a derived ES image (`analysis-icu`) delivered via `canasta image push` + `CANASTA_ELASTICSEARCH_IMAGE`; stock ES crash-loops identically under these probe settings. Live validation of this fix on the same host/cluster is completing now — result will be posted as a comment.

`make validate` and all 1640 unit tests pass; helm template changes only.

### DIFF
--- a/roles/orchestrator/files/helm/canasta/templates/statefulset-elasticsearch.yaml
+++ b/roles/orchestrator/files/helm/canasta/templates/statefulset-elasticsearch.yaml
@@ -42,11 +42,20 @@ spec:
           resources:
             {{- toYaml . | nindent 12 }}
           {{- end }}
+          # Elasticsearch under the CPU limit can take minutes to open
+          # its port on first start; the liveness probe alone would kill
+          # it at ~90s forever. The startup probe owns that window (up
+          # to 5 minutes) and liveness takes over only once it passes.
+          startupProbe:
+            httpGet:
+              path: /_cluster/health
+              port: http
+            periodSeconds: 10
+            failureThreshold: 30
           readinessProbe:
             httpGet:
               path: /_cluster/health
               port: http
-            initialDelaySeconds: 30
             periodSeconds: 10
           livenessProbe:
             httpGet:


cicalese:
Live validation passed on the same host/cluster where the crash loop was observed: with this probe config, the elasticsearch pod (running a derived analysis-icu image from the in-cluster registry) reached 1/1 Ready in ~2 minutes with zero restarts, and serves `/_cat/plugins` → `analysis-icu 7.17.9`. Under the old config the identical setup was killed at ~90s seven times in a row without ever becoming Ready.

PR #1073, opened 2026-07-07
Point a STOPPED Kubernetes instance at canasta start

### LINKED ISSUE #1051: k8s restart interrupted between stop and start leaves the instance scaled to zero
Observed in the CirrusSearch-on-k8s e2e: an SSH connection reset during a `canasta config set` (which restarts to apply) landed between the restart's two halves — after the stop had scaled every workload to 0 and run the user-dir sync, before the start scaled anything back up. The result is a fully-down instance (all deployments/statefulsets at 0 replicas) that `canasta list` still shows as its prior state. Recovery is simply `canasta start`, but nothing tells the operator that.

The long-op detachment added in 4.2.0 protects the individual steps (helm, kubectl waits) from client resets, but the stop→sync→start composite is driven from the controller, so a reset between steps strands the instance at the scaled-to-zero midpoint.

Possible directions: drive the whole restart sequence from a single detached unit the way long helm operations are handled; or make the failure legible — on reconnect/rerun, detect the half-restarted state (workloads at 0 with no stop recorded) and suggest `canasta start`.

cicalese:
Closes #1051.

A Kubernetes restart runs stop (scale all workloads to zero) then start. An SSH reset or other interruption between the two strands the instance at zero replicas, where it reads as STOPPED. Recovery is simply `canasta start`, but nothing told the operator that.

That stranded end-state is byte-for-byte identical to a deliberate `canasta stop` at the cluster level (all workloads at zero, Argo CD auto-sync suspended), so there is no way to single out "stranded restart" without false-positiving on every intentional stop. Instead, `canasta status` now notes the recovery — `canasta start` — whenever a Kubernetes instance reads STOPPED, and flags that an interrupted restart can leave it that way.

Adds unit coverage for the hint appearing when STOPPED and being absent when RUNNING.

### DIFF
--- a/direct_commands/info.py
+++ b/direct_commands/info.py
@@ -380,6 +380,16 @@ def cmd_status(args):
     if orchestrator in ("kubernetes", "k8s"):
         running = _helpers._check_running_k8s(inst_id, host)
         print("Status:       %s" % ("RUNNING" if running else "STOPPED"))
+        if not running:
+            # A restart runs stop (scale to zero) then start; if it is
+            # interrupted in between, the instance is stranded at zero
+            # replicas and reads as STOPPED. That end-state is identical
+            # to a deliberate stop, so point either way at the recovery.
+            print(
+                "              (workloads are scaled to zero; run "
+                "`canasta start --id %s` to bring it up. An interrupted "
+                "restart can leave an instance in this state.)" % inst_id
+            )
         ns = "canasta-%s" % inst_id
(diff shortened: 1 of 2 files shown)


cicalese:
Closing — the stranded state already reports as STOPPED, and an operator who wants the instance up will run `canasta start` without a prompt. The hint adds noise to every STOPPED status for no real gain.

PR #1107, opened 2026-07-09
gitops: ignore legacy config/.smw.json orphan

### LINKED ISSUE #1102: gitops: legacy config/.smw.json path not gitignored, leaks as tracked churn
`.smw.json` is Semantic MediaWiki runtime state. Since CanastaBase 1.2.1 it is written to `config/persistent/.smw.json`, which is already covered by `roles/gitops/files/gitignore.default` (the `config/persistent/` rule).

Instances first initialized under a pre-1.2.1 image have a leftover `config/.smw.json` at the config root — a path no ignore rule covers. Because `gitops init` runs `git add -A`, that orphan is tracked and shows as permanent churn in `gitops status` even on instances now running current images (observed on a farm running canasta:3.5.15).

### Fix
- Add the legacy `config/.smw.json` path to `roles/gitops/files/gitignore.default`.

### Operator note
gitignore alone does not untrack an already-tracked file. Affected instances also need a one-time cleanup:

```
git rm --cached config/.smw.json
rm config/.smw.json
```

SMW recreates it under `config/persistent/` on demand.

cicalese:
`.smw.json` is Semantic MediaWiki runtime state. Since CanastaBase 1.2.1 it is written to `config/persistent/.smw.json`, already covered by the `config/persistent/` rule in `gitignore.default`. Instances first initialized under an older image carry an orphan `config/.smw.json` at the config root — a path no ignore rule covered — so `gitops init`'s `git add -A` tracked it and it shows as permanent churn in `gitops status`.

This ignores the legacy path too.

Operators with an existing orphan also need a one-time `git rm --cached config/.smw.json` (SMW recreates it under `config/persistent/` on demand).

Fixes #1102

### DIFF
--- a/roles/gitops/files/gitignore.default
+++ b/roles/gitops/files/gitignore.default
@@ -33,6 +33,10 @@ images/
 # Runtime state (e.g. Semantic MediaWiki rebuild status)
 config/persistent/
 
+# Legacy SMW state location (pre-CanastaBase 1.2.1, before .smw.json moved
+# to config/persistent/); ignore any orphan left on migrated instances
+config/.smw.json
+
 # Editor / tooling backup files
 *.bak
 *.bak.*


cicalese:
Closing: not a CLI bug. The config/.smw.json churn on the reporting instance was caused by a wiki Settings.php hardcoding $smwgConfigFileDir to the config root, overriding CanastaDefaultSettings.php's $MW_VOLUME/config/persistent default. Removing that override lets SMW write .smw.json to config/persistent/ (already gitignored), so the churn disappears at the source. A correctly-configured instance never creates config/.smw.json, so a defensive ignore rule would guard a scenario that shouldn't occur.

PR #1121, opened 2026-07-11
Remove depends_on: db from web service (fixes #1119)

### LINKED ISSUE #1119: 06 depends_on db with required:false hangs podman-compose 1.6
## Summary
The Canasta compose template (`roles/orchestrator/files/compose/docker-compose.yml`)
defines the `web` service with:

```yaml
depends_on:
  db:
    condition: service_healthy
    required: false
```

The `condition: service_healthy` + `required: false` syntax is Docker
Compose v2 (Specification) format. **podman-compose 1.6.0** does not
understand `required: false` and **hangs indefinitely** waiting for the
`db` service healthcheck — even though `db` is profile-gated
(`profiles: [internal-db]`) and never starts.

## Environment
- Debian 12 Bookworm, r7g.large EC2 (ARM64)
- Podman 4.3.1 (rootless)
- podman-compose 1.6.0 (pip-installed)
- Canasta CLI v4.12.0
- Canasta image ghcr.io/canastawiki/canasta:3.5.15

## Observed
`canasta create` calls `docker compose up -d` (via podman-compose shim).
Only the `caddy` container starts (it has no `depends_on`). The `web`
container is never created — podman-compose blocks waiting for the `db`
healthcheck. The command hangs indefinitely (observed > 1 hour) with no
output or error.

## Root Cause
podman-compose 1.6.0 parses the `depends_on` block and treats
`required: false` as an unknown field. It then enters the normal
dependency-wait path for `db: condition: service_healthy`, but since `db`
is behind a profile that is never activated, the wait never completes.

In Docker Compose v2, `required: false` means "do not fail if the
dependency service is not present (e.g., due to profiles)." podman-compose
1.6.0 does not implement this flag.

## Workaround Applied
We patch the compose template before `canasta create` to remove the
entire `depends_on` block for `db`:

```bash
sudo sed -i '/^    depends_on:/{N;N;N;/required: false/d}' "$COMPOSE_TPL"
```

This removes only the `db` dependency block (other `depends_on` blocks
use the list form and don't contain `required: false`, so they survive).

## Suggested Fix
1. Remove `required: false` from the `depends_on` block — it is
   redundant since `db` is already profile-gated. When `db` is not
   started (no profile), Docker Compose v2 simply skips the dependency
   check without `required: false`.
2. If `required: false` must stay for some reason, document a minimum
   podman-compose version (2.0+?) that supports it.
3. Alternatively, restructure as a list-form `depends_on` which
   podman-compose 1.6.0 supports:

```yaml
# List form (no condition/required) — works on both Docker and podman
depends_on:
  - db
```

hexmode:
## Summary

The compose template's `web` service had a `depends_on: db` block using Docker Compose v2 syntax (`condition: service_healthy` + `required: false`). **podman-compose 1.6.0** does not understand `required: false` and hangs indefinitely waiting for the `db` service healthcheck — even though `db` is profile-gated (`profiles: [internal-db]`) and never starts.

## Why this is safe to remove

The `required: false` field is redundant: when `db` is not started (no profile activated), Docker Compose v2 already skips the dependency check without it. The `db` service is exclusively used when `USE_EXTERNAL_DB=false` (internal database), at which point it's started via the `internal-db` profile and the dependency is satisfied naturally.

Removing the entire `depends_on` block is safe for both Docker and podman.

## Testing

- Tested with podman-compose 1.6.0 + Podman 4.3.1: `canasta create` completes successfully, all containers start (web, caddy, varnish)
- Tested with Docker Compose v2: no change in behavior (db is profile-gated, never started without the profile)

Fixes #1119


cicalese:
Confirming this is safe from a startup-ordering standpoint: the web container already waits for the database itself (`roles/mediawiki/tasks/wait_for_db.yml` + `canasta_db_wait_timeout`, plus the entrypoint's wait-for-it), so even in the internal-db case (`USE_EXTERNAL_DB=false`, `internal-db` profile active) there's no startup race introduced by dropping the compose-level `depends_on`. The gate was redundant with the app-level DB wait.

PR #1122, opened 2026-07-11
feat: native Podman support (fixes #1118)

### LINKED ISSUE #1118: 07 canasta create shells out to docker compose — no native podman support
## Summary
Canasta CLI shells out to `docker compose` for all container
orchestration. On systems using rootless Podman instead of Docker,
this requires a custom `/usr/bin/docker` wrapper script that routes
`docker compose` → `podman-compose` and handles subtle differences
(e.g., `docker ps` output format, `docker inspect` multi-ID arguments).

This wrapper is fragile and needs maintenance with every Canasta or
podman-compose update.

## Environment
- Debian 12 Bookworm, ARM64 EC2
- Podman 4.3.1 (rootless), podman-compose 1.6.0
- Canasta CLI v4.12.0

## Observed
Canasta CLI hardcodes `docker compose` commands (e.g., in
`roles/create/tasks/main.yml`). Without a `docker` binary, `canasta
create` fails immediately. We created a 30-line `/usr/bin/docker`
wrapper that:
- Routes `docker compose <args>` → `podman-compose -f docker-compose.yml <args>`
- Routes `docker ps` → `podman ps` with post-processing to match Docker's format
- Routes `docker inspect` → a custom wrapper that splits space-separated container IDs

## Workaround Applied
A custom `/usr/bin/docker` wrapper script (installed by user-data.sh
during cloud-init). This works but is fragile — any change to
podman-compose's output format or Canasta's expected `docker` behavior
breaks the wrapper silently.

## Suggested Fix
1. Add a `--orchestrator=podman` flag (or auto-detect) that uses
   `podman-compose` directly instead of `docker compose`.
2. Alternatively, use the Python `compose` library or a runtime-agnostic
   abstraction instead of shelling out to a specific CLI.
3. At minimum, document the podman deployment path and the required
   wrapper script in the setup guide.

### LINKED ISSUE #1158: canasta list crashes with TypeError when wiki URL is null
## Description

`canasta list` crashes with a TypeError when a wiki entry has a null/empty URL in the registry.

## Steps to Reproduce

1. Have a Canasta instance with a wiki entry that has no URL set (null value in YAML)
2. Run `canasta list`

## Error

```
  File "/opt/canasta-ansible/canasta.py", line 1708, in <module>
    main()
  File "/opt/canasta-ansible/canasta.py", line 1648, in main
    result = direct_commands.run_direct_command(command_name, args)
  File "/opt/canasta-ansible/direct_commands/_helpers.py", line 37, in run_direct_command
    return handler(args)
  File "/opt/canasta-ansible/direct_commands/info.py", line 141, in cmd_list
    _helpers._print_table(details)
  File "/opt/canasta-ansible/direct_commands/_helpers.py", line 1155, in _print_table
    if "/" not in url,
TypeError: argument of type 'NoneType' is not iterable
```

## Cause

`w.get("url", "")` returns `None` when the YAML key exists but the value is null. The `in` operator fails on `None`.

## Fix

Change `w.get("url", "")` to `w.get("url") or "" ` in `_helpers.py` line 1154.

hexmode:
## Summary

Full Podman support across all Canasta-CLI paths. Also fixes #1158 (canasta list crash on null URL).

### Registry and Config
- Add compose_command/inspect_command fields to registry and defaults
- .env fallback when registry lacks compose fields (legacy instances)
- Always set compose_command/inspect_command in extra_vars to shield from Ansible role variable scoping

### Auto-Detection
- Auto-detect Podman-only hosts in canasta.py (no Docker binary present)
- Auto-detect Podman as container engine in canasta-docker via CONTAINER_ENGINE env var
- Detect podman socket path when CONTAINER_ENGINE=podman
- Skip local auto-detection when --host is set (remote host may have different runtime)

### Ansible Roles
- Use compose_command variable in create, preflight, copy_into_container, sync_compose_profiles, and create.yml rescue block
- Accept podman info in preflight when compose_command contains podman
- Independent runtime detection: probe Docker then Podman, set facts if only Podman available

### Podman-compose Compatibility
- Replace compose ps -q web with native docker ps --filter label=com.docker.compose.service=web (podman-compose ps does not accept service name args)
- Same fix for ps --services --filter and ps web --format in other tasks

### Bug Fixes
- canasta list: handle null wiki URL (#1158)
- canasta doctor checks podman version, warns if < 4.4

## Known Issues
- podman-compose DNS resolution between containers (service names) - filed as #1160


cicalese:
Opened #1139 to track the full runtime-parameterization effort — this PR is a good first step, but it's step 1 of ~25 call sites, so flagging the scope here:

- **Reach of this change:** `canasta start`/`stop`/`restart` on Compose normally run the Python fast-path in `direct_commands/lifecycle.py` (→ `_helpers._run_compose`), which hardcodes `["docker", "compose"]`. Ansible `start.yml` only runs for K8s or sidecar instances, so a sidecar-less Compose instance would start under Podman via `create` but still `stop`/exec/backup under `docker`. Full call-site list and a single-source-of-truth proposal are in #1139.
- **In-diff fix:** the `start.yml` comment suggesting `compose_command: "podman-compose -f docker-compose.yml"` duplicates the base file — `_compose_files` already begins with `-f docker-compose.yml`, so this yields `... -f docker-compose.yml -f docker-compose.yml ... up -d`. The value should be just `podman-compose` (matching the `defaults/main.yml` comment).

cicalese (changes requested):
Requesting changes — the direction is right, but as-is the advertised Podman start doesn't take effect for the path users hit, and the setup guidance in the diff is wrong. Full scope and a single-source-of-truth proposal are in #1139; summarizing the asks for this PR:

1. **Fast-path reach.** `canasta start`/`stop`/`restart` for sidecar-less Compose instances run `_helpers._run_compose` (hardcoded `docker compose`), so `compose_command` is ignored except during `create` and sidecar instances. Either scope this PR down (land the plumbing, don't document Podman as working yet) or extend the override to the Python fast-path — see #1139.
2. **In-diff bug.** The `start.yml` comment `compose_command: "podman-compose -f docker-compose.yml"` duplicates the base file — `_compose_files` already begins with `-f docker-compose.yml`. It should be just `podman-compose`.

Nits:
- The `| default('docker compose')` / `| default('docker')` in the `set_fact` is dead — both vars are defined in `defaults/main.yml` and can't be undefined.
- Please note which Podman version you validated `inspect -f '{{.State.Health.Status}}'` against; the PR asserts docker/podman parity for the health templates without a test.

cicalese (changes requested):
Thanks for the rework — the fast-path reach issue from my last review is properly fixed now: `_resolve_compose_cmd` in `direct_commands/_helpers.py` covers `_run_compose`, `_exec_in_container`, `_stream_in_container`, `rebuild`, and `info`, which was the main gap. The `-f docker-compose.yml` duplication in the `start.yml` comment is fixed too.

I'm still requesting changes. The blockers this round are mostly **Docker regressions**, not Podman gaps — they wouldn't show up in single-instance Podman testing. `make test-unit` (1851), `make validate`, and `make lint` all pass on the branch, so none of these are caught by CI.

## Blocking

### 1. `_cscli_argv` is broken for Docker Compose — all CrowdSec commands fail

`roles/crowdsec/tasks/_resolve_cscli.yml`:

```yaml
_cscli_argv:
  - "{{ compose_command }}"
```

`compose_command` defaults to the string `"docker compose"`, so this produces a **single** argv element:

```
["docker compose", "exec", "-T", "crowdsec", "cscli", ...]
```

`ansible.builtin.command` with `argv:` execs directly without a shell, so this looks for a binary literally named `docker compose`. Every consumer breaks — `alerts.yml`, `ban.yml`, `unban.yml`, `console_enroll.yml`, `reload.yml` — on plain Docker, not just Podman. The previous value was `['docker', 'compose', 'exec', '-T', 'crowdsec']`.

Fix: `_cscli_argv: "{{ compose_command.split() }}"`.

`test_crowdsec.py::test_compose_prefix` was updated to assert on `_cscli_prefix` (the string form) only, so the argv breakage passes CI.

### 2. `ps --filter label=...` is not project-scoped — breaks multi-instance hosts

Every replacement of `docker compose ps` with `docker ps --filter label=com.docker.compose.service=web` drops project scoping. `docker ps` is a daemon-level query — `chdir` / `cwd` has no effect on it. I confirmed this locally: running that exact command from the Canasta-CLI repo (not an instance dir) returns the `web` container of my `dev` instance.

Affected (all of these run for Docker users too, not just Podman):

- `direct_commands/_helpers.py::_check_running_compose` → `canasta list`, `canasta status`
- `direct_commands/_helpers.py::_gather_instance_info` (remote branch)
- `roles/orchestrator/tasks/check_running.yml`
- `roles/orchestrator/tasks/list_running_services.yml`
- `roles/orchestrator/tasks/start.yml` ("Get web container id")
- `roles/orchestrator/tasks/upgrade_rebuild_buildable.yml` ("Get running web container image ID")
- `roles/crowdsec/tasks/_preflight.yml`

Concrete failures with two or more instances on one host (the documented model):

- `canasta list` reports every stopped instance as RUNNING as soon as any one instance is up.
- `start.yml`: `_start_web_cid.stdout` is now multi-line, and it's passed as a *single* argv element to `inspect -f`. The health gate then either inspects a foreign container or spins for the full 60 × 10s = 10 min before failing.
- `upgrade`: the "running image" is read from whichever instance `docker ps` happened to list first.
- `list_running_services` / crowdsec preflight report another instance's services.

The label filter needs a project predicate: `--filter label=com.docker.compose.project=<project>`. Nothing currently sets `COMPOSE_PROJECT_NAME` (I grepped — no hits in the repo), so the project name is the sanitized directory basename, which is not always the instance id when `-p` was used. Two options:

- Write `COMPOSE_PROJECT_NAME={{ id }}` into `.env` at create time and filter on it (also makes volume naming deterministic), or
- Keep `docker compose ps` on the Docker path and use the label form only when `compose_command` is Podman-based.

The second is the smaller change and keeps Docker behavior byte-identical.

### 3. `create_preflight.yml` silently sudo-edits the host's sysctl config

```yaml
- name: Raise unprivileged port floor to 80 for rootless runtimes
  when: ansible_user_id != 'root'
  become: yes
  block:
    - ansible.builtin.slurp: {src: /proc/sys/net/ipv4/ip_unprivileged_port_start}
    - ansible.posix.sysctl:
        name: net.ipv4.ip_unprivileged_port_start
        value: '80'
        sysctl_set: yes
        reload: yes
```

Four problems:

- It's at column 0 — **outside** the `when: orchestrator | default('compose') == 'compose'` block — so it runs on Kubernetes creates as well, where it's irrelevant.
- `become: yes` escalates to root during `canasta create`. On any host without passwordless sudo this aborts the create; nothing in the command's contract says it needs root.
- `ansible.posix.sysctl` with `sysctl_set: yes` + `reload: yes` and the default `sysctl_file` **persistently rewrites `/etc/sysctl.conf`**. That's a permanent, host-wide kernel setting change the operator never asked for.
- No `ansible_system == 'Linux'` guard and no `failed_when`. The `slurp` of `/proc/sys/...` fails outright on a macOS target.

`canasta doctor` already detects this condition and prints the exact two sudo commands for the operator to run (see the "Privileged ports: BLOCKED" branch in `direct_commands/doctor.py`). That's the right division of labor — please drop this block and lean on doctor, or at minimum gate it behind an explicit opt-in flag, restrict it to `orchestrator == 'compose'` and `ansible_system == 'Linux'`, and don't persist it to `/etc/sysctl.conf`.

### 4. Agent scaffolding and a personal debug script are committed

`.sisyphus/` is ~1,200 lines of session state, QA logs, plan files, and run-continuation JSON, with absolute paths (`/home/mah/work/code/Canasta-CLI`) and session IDs. `tmp-run` is a personal one-off shell script, also with `/home/mah` paths baked in.

Neither is gitignored (`git check-ignore` returns nothing). Please drop both from the PR.

## High

### 5. `canasta doctor` reports Podman on hosts that don't have it, and its dependency gate no longer fails

The probe is `podman --version 2>&1 || echo MISSING`. When podman is absent, the shell's own "not found" message is captured by the `2>&1` and lands on stdout, so the field is:

```
sh: podman: command not found
MISSING
```

I verified this. Two consequences:

- `_parse_doctor` tests `podman_version != "MISSING"`, which is now always true → every Docker-only host gets a spurious `Podman:  sh: podman: command not found MISSING (version unknown)` line in the report.
- Worse, `cmd_doctor`'s core-dependency gate:
  ```python
  elif podman_version != "MISSING":
      pass
  ```
  means a host with **neither Docker nor Podman** now passes `canasta doctor` with rc 0. That gate was the one hard failure in the command.

The existing probes avoid this by testing for a substring of the expected output (`"Docker" in docker`) rather than for the sentinel. Suggest `command -v podman >/dev/null 2>&1 && podman --version 2>/dev/null || echo MISSING` plus a `"podman" in podman_version.lower()` test.

### 6. `compose_command` / `inspect_command` are undefined for `--host` operations outside the orchestrator role

`canasta.py` skips the injection when `--host` is set (`if "compose_command" not in extra_vars and not host_value:`), and the `.env` fallback above it reads a *local* path, which doesn't exist for a remote instance. So neither var gets set.

The defaults live in `roles/orchestrator/defaults/main.yml`, but these playbooks `include_role` a different role entirely and never pull orchestrator in:

- `roles/devmode/tasks/enable.yml` — bare `{{ inspect_command }}` (7 uses) and `{{ compose_command }}`
- `roles/upgrade/tasks/migrations/mysql_to_mariadb.yml` — bare `{{ inspect_command }}` (10 uses), `{{ compose_command }}`
- `roles/crowdsec/tasks/_resolve_cscli.yml`, `_restart_engine.yml` — bare `{{ compose_command }}`, plus `when: compose_command is search('podman')` which raises on an undefined var

So `canasta crowdsec ban -H host`, `canasta devmode enable -H host`, and the MySQL→MariaDB migration path of `canasta upgrade -H host` all fail with an undefined-variable error. (`roles/create/tasks/_validate_inputs.yml` got this right with `| default('docker')`.)

Cleanest fix is to define both at play scope in `canasta.yml` rather than in one role's defaults, then the extra-vars shim becomes an override rather than a load-bearing requirement.

### 7. The controller's local runtime is pinned onto remote hosts, and the remote probe can't override it

The `not host_value` guard only covers the explicit `-H` form. For a **registry-resolved remote instance** — the common case, no `-H` on the command line — `host_value` is `None`, so the auto-detect fires and injects the *controller's* runtime as an extra var, which is then applied to the remote target.

Because extra-vars outrank `set_fact`, `create_preflight.yml`'s "Set compose_command to podman-compose if Docker absent but Podman present" can never take effect once `canasta.py` has set it. That also kills the fallback in a plain local case: Docker binary present but daemon down, Podman running → extra-vars pins `docker compose`, the preflight gate passes (podman probe succeeded), and the create then fails at `up -d`.

Since `_register.yml` now persists `composeCommand`/`inspectCommand` to the registry, the resolution order should be registry → `.env` → remote probe, with the controller's PATH used only for a genuinely local, unregistered create.

## Medium

8. **`parts` shadowing** — `direct_commands/doctor.py`, the podman version block does `parts = [int(x) for x in ver_str.split(".")]`, rebinding the `parts` that the `p(i)` closure reads. Harmless today because no `p()` call follows, but the file's whole convention is "append the next probe at the end and index it," so the next person to add one gets a `TypeError`. Rename to `ver_parts`.

9. **`CONTAINER_ENGINE` isn't honored.** The PR body says "auto-detect Podman as container engine in canasta-docker via `CONTAINER_ENGINE` env var," but `canasta-docker` assigns it unconditionally, so `CONTAINER_ENGINE=podman ./canasta-docker ...` is ignored and there's no way to pick Podman on a host that has both. `CONTAINER_ENGINE="${CONTAINER_ENGINE:-...}"` around the detection.

10. **MYSQL defaults in `_env_update.yml` apply to Docker instances too.** The values do match the `${MYSQL_HOST:-db}` / `:-3306` / `:-root` / `:-false` defaults in `docker-compose.yml`, so this is semantically correct — but it adds four new keys to `.env` (and therefore to `env.template` and every gitops diff) for every new Docker instance, to work around a podman-compose interpolation limitation. Please gate it on `compose_command is search('podman')`.

11. **Dockerfile collection versions duplicated.** `K8S_CORE_VERSION`/`ANSIBLE_POSIX_VERSION` are hardcoded in the Dockerfile with a comment saying they "must match requirements.yml," but nothing enforces it. This repo tests this kind of invariant elsewhere (e.g. the Elasticsearch parity guard) — a unit test asserting the two agree would prevent silent drift.

12. **The requirements lockfile is a separate PR.** `requirements.in` + a 177-line pinned `requirements.txt` is a defensible change on its own merits, but it's unrelated to Podman and it bumps the lint toolchain (`ansible-lint` 26.6.0, `ruff` 0.16.0, `yamllint` 1.38.0), which changes CI behavior for everyone. Also, `pip install --only-binary :all:` will hard-fail on any dependency that ships sdist-only for the build arch — has this been exercised on both amd64 and arm64? Same for the `kubernetes.core` `>=6.4.0` → exact `6.5.0` pin, which required loosening `test_kubernetes_structure.py` with a `pin == "6.5.0"` special case that weakens the test's stated intent.

13. **`<<: *default-network` in `docker-compose.yml`.** This adds an explicit `networks: default: {driver: bridge}` and a merge key on all 11 services, for a podman-compose DNS problem that the PR's own "Known Issues" says is still unresolved (#1160). If it doesn't fix anything, please drop it rather than carry it for all Docker users.

14. **No tests for the new Podman logic.** Nothing covers `_resolve_compose_cmd`, `_resolve_inspect_cmd`, `_compose_profile_args`, `_read_env`, the doctor podman version parse, or the `canasta.py` auto-detect. The test changes in this PR are all mechanical updates of existing YAML string assertions to the new Jinja literals — several now assert less than they did (`test_crowdsec.py` went from `"docker compose ps" in preflight` to `"{{ inspect_command | default('docker') }}" in preflight`, which no longer says anything about CrowdSec). Findings 1, 2, 5, and 6 are all unit-testable.

15. **`_read_env` is duplicated** — `_read_env_file_value` in `canasta.py` and `_read_env` in `direct_commands/_helpers.py` are byte-identical. Neither strips surrounding quotes, so `compose_command="podman-compose"` in a hand-edited `.env` yields a bogus token.

16. **`info.py::cmd_status`** resolves `compose_cmd` and then calls `_helpers._check_running_compose(path, host)` without passing it.

17. **`podman ps --format '{{.Label "..."}}'`** — used in `list_running_services.yml` and `crowdsec/_preflight.yml`. Podman's `ps` template context exposes `Labels` as a map; I don't have a Podman host to check whether the Docker-compat `.Label "key"` *function* is supported there. Could you confirm which Podman version you validated that format string against?

## Question about what was validated

The PR's "Known Issues" says podman-compose inter-container DNS (service-name resolution) doesn't work, filed as #1160. If `web` can't resolve `db` and `caddy` can't resolve `web`, that's not a rough edge — the wiki doesn't come up. So I want to make sure I understand what "successfully tested" covers here:

- Podman with its **Docker-compatible socket** (`podman.sock` + real `docker compose`), which is the `DOCKER_HOST` path that already worked, or
- **podman-compose**, the separate Python tool, which is what this PR's `compose_command` plumbing actually targets?

Those are different products with different flag surfaces, and the answer changes what this PR should claim. If it's the first, the podman-compose paths here are untested and the PR should be scoped to the socket route. If it's the second, could you post the `canasta create` → browse-the-wiki transcript, since that contradicts #1160?

Either way, I'd suggest not advertising "full Podman support" in the release notes until #1160 is closed — landing the plumbing with a documented "experimental, Podman ≥ 4.4 via the Docker socket" note is fine and gets the refactor in.

## Stale item in the description

The PR body says "Also fixes #1158 (canasta list crash on null URL)," but there's no such change in the diff — #1158 was fixed separately and closed on 2026-07-24. Worth dropping from the description so the merge doesn't auto-close anything unexpectedly.

## Suggested split

At 2,708 additions across 57 files this is hard to reason about as one unit, and the findings above cluster cleanly into independent pieces. Reviewing and reverting get much easier if this lands as a sequence. Proposed order — each is independently useful and independently revertable:

**PR 1 — build reproducibility** (no behavior change; issue needed)
`requirements.in` + pinned `requirements.txt`, the curl-based galaxy collection install, `ansible.posix` added to `requirements.yml`, `--root-user-action=ignore`.
Commits roughly `89bb631`, `7c59bfd`, `bbb9b50`. Open questions to settle here: `--only-binary :all:` on both amd64 and arm64, and a test asserting the Dockerfile's pinned collection versions match `requirements.yml` (finding 11). This unblocks your local builds and is reviewable in isolation.

**PR 2 — runtime parameterization** (#1139)
`compose_command`/`inspect_command` end to end: registry fields, `defaults`, the `_resolve_compose_cmd`/`_resolve_inspect_cmd` helpers, and the substitution across roles and `direct_commands/`. Defaults stay Docker, so this is a pure refactor with no user-visible change.
Commits roughly `99d1fa2`, `0d63766`, `477d154`, `d4ff219`.
Must include fixes for findings 1 (`_cscli_argv` split), 6 (play-scope the defaults instead of relying on the extra-vars shim), and unit tests. This is the bulk of the value and the piece most worth getting right — everything else layers on top.

**PR 3 — container-state queries** (needs its own issue)
The `compose ps` → `docker ps --filter label=...` conversions, with project scoping added (finding 2). Small, self-contained, and the one change most likely to need a revert if the project-name derivation turns out wrong — much safer on its own than buried in a 57-file diff.

**PR 4 — Podman runtime detection and selection** (#1118)
`canasta-docker` engine selection + socket detection, the `canasta.py` auto-detect, `create_preflight.yml`'s dual probe, and the doctor podman check. Findings 5, 7, 8, 9 live here.
Commits roughly `21bb8e1`, `0303d94`, `971c3a5`, `c71a9d3`, `64462cb`, `a0049e2`, `0739e30`.

**PR 5 — podman-compose compatibility** (#1118)
Explicit `--profile` flags, the `.env` MySQL defaults gated on Podman (finding 10), and the network stanza only if it turns out to matter for #1160.
Commits roughly `b2eaf18`, `76925bc`, `7461da6`, `5f414f3`.

**Dropped, not split:** `.sisyphus/`, `tmp-run`, and the sysctl block (finding 3). If auto-raising the port floor is something you want, it deserves its own issue and design discussion — an opt-in flag, Linux-only, non-persistent — rather than riding along in a Podman PR.

PRs 1–3 are all Docker-only and could go in well ahead of the Podman work, which would also shrink what the multi-node release validation has to cover for 4 and 5.

Happy to help with the mechanics if that's useful — I can open the issues for the pieces that don't have one yet (PRs 1 and 3), and rebase-split the branch into the five stacked branches for you to review and push under your own name. Just say which parts you want to keep driving yourself.

cicalese:
@hexmode — split into five sequenced PRs as discussed. Every commit carrying your work is authored to you; the two commits that aren't yours are called out below. Keeping this PR open as the tracking PR for the series.

Each branch is stacked on the previous one, so review in order — each PR's diff is only meaningful on top of its base:

| PR | Base | Size | Issue |
|---|---|---|---|
| #1223 build: pin requirements, curl collection installs | `main` | 5 files, +211/−14 | #1221 |
| #1224 refactor: parameterize the container runtime | #1223 | 30 files, +309/−109 | #1139 |
| #1225 fix: query container state via compose labels | #1224 | 6 files, +21/−8 | #1222 |
| #1226 feat: detect Podman as the container runtime | #1225 | 4 files, +144/−32 | #1118 |
| #1227 feat: compose profiles + DB defaults for podman-compose | #1226 | 15 files, +311/−24 | #1118 |

`make test-unit`, `make lint`, and `make validate` all pass on each branch independently, not just at the tip.

**Verification that nothing was lost:** diffing the top of the stack against this branch shows only the intended differences — `.sisyphus/`, `tmp-run`, the sysctl block, and the two files touched by the fix below. No part of your work was silently dropped.

**Two commits are not yours.** In #1224 I fixed `_cscli_argv` (`compose_command.split()`) and added a regression test, because that PR's premise is "no behavior change" and shipping the crowdsec breakage would have made that false. Both are separate commits authored to me, not folded into yours.

**Deliberately not carried over**, each explained in the relevant PR:

- `.sisyphus/` and `tmp-run` — looked like local agent scaffolding and a personal debug script, both with `/home/mah` paths baked in
- the `ip_unprivileged_port_start` sysctl block — needs its own issue and design pass (it escalated with `become: yes` during `create`, ran for Kubernetes too, and persisted to `/etc/sysctl.conf`)

**Everything else from the review stays open as review items on the PR it belongs to**, rather than being fixed by me — several need a design call that's yours to make. The two biggest: project-scoping the label filter (#1225, written up in #1222) and the resolution order for remote hosts (#1226).

One thing that would help most: the question at the bottom of #1227 about whether the validated setup was podman-compose or Podman's Docker-compatible socket. That determines how much of #1225 and #1227 is actually exercised.

Also — this PR's description says it fixes #1158, but that was fixed separately and closed on 2026-07-24, and there's no such change in the diff. Worth dropping from the description.

Happy to rework any of the splits if the boundaries don't match how you'd rather sequence it.

cicalese:
@hexmode — one focused question, separated out from my longer comments because the answer determines how much of #1225 and #1227 your testing actually exercised.

When you validated this, which setup was it?

**A. podman-compose** — the separate Python tool, invoked as `podman-compose ...`. This is what the `compose_command` plumbing targets, and it's why #1227 exists at all: the explicit `--profile` flags are there because podman-compose ignores `COMPOSE_PROFILES` from `.env`, and the `.env` MySQL defaults are there because it can't expand `${VAR:-default}`.

**B. Podman's Docker-compatible socket** — `podman.sock` with real `docker compose` on top, `DOCKER_HOST` pointed at it. Here `compose_command` stays `docker compose` and none of the podman-compose workarounds fire.

Quickest way to tell, on the instance you tested:

```bash
grep -E '^(compose_command|inspect_command)=' /path/to/instance/.env
```

Plus whatever `DOCKER_HOST` was set to, and the `composeCommand` field for that instance in `conf.json` if there is one.

Why I'm asking: this PR's Known Issues lists podman-compose inter-container DNS as still unresolved (#1160). If service names don't resolve, `web` can't reach `db` and `caddy` can't reach `web` — a wiki shouldn't come up at all under A. That's hard to square with a successful end-to-end test, so I suspect B, in which case the socket route is the validated one and the podman-compose paths in #1225 and #1227 are still unexercised.

B would be a perfectly good result, to be clear — it just changes what the series should claim and probably means stopping at #1226 until #1160 is sorted. And if it really was A, a `canasta create` → browse-the-wiki transcript would be genuinely useful, because it would mean #1160 is stale.

cicalese:
Closing — the whole series has landed. Thanks @hexmode; every commit carrying your work kept your authorship through the split.

| PR | | status |
|---|---|---|
| #1223 | build reproducibility (#1221) | merged |
| #1224 | runtime parameterization (#1139) | merged |
| #1225 | container-state queries (#1222) | merged |
| #1226 | Podman runtime detection (#1118) | merged |
| #1227 | podman-compose profiles + DB defaults (#1118) | merged |

Two things from this branch were deliberately left out, and neither is lost — they just need their own issues rather than a stale PR carrying them:

- **The `ip_unprivileged_port_start` sysctl block.** It ran at top level (so on Kubernetes creates too), escalated with `become: yes` during `canasta create`, and persisted to `/etc/sysctl.conf` via `sysctl_set` + `reload`, with no `ansible_system == 'Linux'` guard. `canasta doctor` already detects the condition and prints the exact commands. If we want to automate it, it deserves an opt-in flag and proper guards.
- **`.sisyphus/` and `tmp-run`** — local scaffolding and a personal debug script.

Worth recording what the split turned up, since none of it was visible from the PR checks:

- `_cscli_argv` built `docker compose` as a single argv element, breaking every crowdsec subcommand on **plain Docker**.
- The runtime substitution missed twelve executed call sites, including `canasta exec` and the restic backup invocation — the latter diverging from its own Ansible equivalent.
- `canasta.py` called an unimported `_helpers`, which a narrowed `except` turned from silently swallowed into a hard `NameError` on every Ansible-dispatched command.
- The doctor podman probe let a host with **no container runtime at all** pass the dependency gate.
- The container-state label filter wasn't project-scoped, so on a multi-instance host `canasta list` reported stopped instances as RUNNING. The GitOps integration job reproduced it — `docker inspect` receiving two newline-joined container IDs — and passed once scoped.

Every one of those was green on all seven PR checks. The integration jobs only run on push-to-main or via `gh workflow run tests.yml --ref <branch>`, which is what caught the last one.

#1118 stays open as the tracking issue for Podman; a hands-on e2e of the merged stack is next.

Issue #1077, opened 2026-07-07
wiki-check: design for user-action checks (edit/delete/upload) — how to authenticate


cicalese:
Follow-up to #569 (closed) and #570. The initial `wiki-check` command verifies wiki **reachability** (HTTP GET to `api.php`, asserting a MediaWiki response). #569 also envisioned checks for actions that cause changes — editing/creating pages, deleting, file uploads. Those require acting as an authenticated user. This issue records the recommended design before that work starts.

## Question

For write checks, does the command need to take a username/password, or is there a built-in identity to reuse?

## What is already built in

- Every wiki is installed with admin user `WikiSysop` (overridable via `--admin`), created by MediaWiki `install.php`.
- Its password is persisted on the host at `<instance_path>/admin-password_<wiki-id>` (mode 0600) and mirrored in GitOps as `admin_password_<wid>`; both are already treated as secrets (git-crypt / gitignored). So the CLI can obtain a working identity without prompting the operator. (Nothing reads that file back today — the plumbing would be new but trivial.)
- The CLI already performs privileged wiki writes with **no HTTP login**, via `php maintenance/run.php <script>` inside the container over SSH/exec (`direct_commands/maintenance.py`; e.g. `createAndPromote`, `update.php`, `runJobs.php`).
- Bot-password precedent exists in-repo: the `wiki-publish` workflow authenticates to the docs wiki with `WIKI_BOT_USER` (`User@BotName`) + `WIKI_BOT_PASSWORD`.

## Two mechanisms, by layer

**A. In-container maintenance scripts — no credentials.** The pattern the CLI already uses. MediaWiki ships scripts that mutate as a named user, bypassing HTTP/auth — e.g. `edit.php --user WikiSysop`, `importImages.php` (uploads), `deleteBatch.php` / `nukePage.php` (deletes). Best for verifying the **application/data layer** (DB writable, parser/extensions work, jobs run). Zero-config and robust, but does **not** exercise Caddy -> web -> auth, so it will not catch web-tier write failures a real editor would hit.

**B. HTTP API as a real user — bot password.** The right way to test the **full user-facing write path** (mirrors what the current read check does over HTTP). Use a scoped **bot password** (`Special:BotPasswords`), never the main account password — revocable, grant-scoped (edit/delete/upload), does not trip 2FA. Credentials source, either:
- operator-supplied flag/env (explicit; requires minting a bot password first), or
[...]

PR #1123, opened 2026-07-12
Implement write checks for wiki-check command

naresh-kumar-babu:
Implements #1077.

This PR implements the new `--write` flag for the `wiki-check` command. It enables full end-to-end mutation and write verification (page creation/editing and file uploading) for MediaWiki instances, ensuring write accessibility is operational.

### DIFF
--- a/direct_commands/_mediawiki_api.py
+++ b/direct_commands/_mediawiki_api.py
@@ -1,0 +1,131 @@
+import http.cookiejar
+import json
+import secrets
+import ssl
+import urllib.parse
+import urllib.request
+
+
+class MediaWikiClient:
+    def __init__(self, api_url, host_header=None):
+        self.api_url = api_url
+        self.host_header = host_header
+        cj = http.cookiejar.CookieJar()
+        handlers = [urllib.request.HTTPCookieProcessor(cj)]
+        if api_url.startswith("https"):
+            context = ssl._create_unverified_context()
+            handlers.append(urllib.request.HTTPSHandler(context=context))
+        self.opener = urllib.request.build_opener(*handlers)
+
+    def _request(self, params=None, files=None):
+        headers = {}
+        if self.host_header:
+            headers["Host"] = self.host_header
+
+        if files:
+            content_type, body = self._encode_multipart(params or {}, files)
+            headers["Content-Type"] = content_type
+            req = urllib.request.Request(self.api_url, data=body, headers=headers)
+        else:
+            data = urllib.parse.urlencode(params or {}).encode("utf-8")
+            req = urllib.request.Request(self.api_url, data=data, headers=headers)
+
+        with self.opener.open(req, timeout=30) as resp:
+            return json.loads(resp.read().decode("utf-8", errors="ignore"))
+
+    def _encode_multipart(self, fields, files):
+        boundary = secrets.token_hex(16)
+        parts = []
+        for name, value in fields.items():
+            parts.append(f'--{boundary}')
+            parts.append(f'Content-Disposition: form-data; name="{name}"')
+            parts.append('')
+            parts.append(str(value))
+        for name, (filename, content, mimetype) in files.items():
+            parts.append(f'--{boundary}')
+            parts.append(f'Content-Disposition: form-data; name="{name}"; filename="{filename}"')
+            parts.append(f'Content-Type: {mimetype}')
+            parts.append('')
+            if isinstance(content, str):
+                content = content.encode("utf-8")
+            parts.append(content)
+        parts.append(f'--{boundary}--')
+        parts.append(b'')
+
+        body = b""
+        for part in parts:
+            if isinstance(part, str):
+                body += part.encode("utf-8") + b"\r\n"
+            else:
+                body += part + b"\r\n"
+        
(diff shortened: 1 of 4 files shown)


yaronkoren:
This looks great! But could/should the MediaWikiClient class be put into its own file?

naresh-kumar-babu:
> This looks great! But could/should the MediaWikiClient class be put into its own file?

Yes, it can. 
Thanks for the suggestion. Updated the PR.

yaronkoren:
Sorry for the delay. One issue here is that the page and file name are too hardcoded - the code should ensure that a page and file with this name don't already exist. One easy way to do that is to add a random number to the end so that, for example, temp_page is not set to "Canasta-Wiki-Check-Temp-Page", but rather "Canasta-Wiki-Check-Temp-Page-XXX", where "XXX" is a random three-digit number. That way the code can cycle through random three-digit numbers until it finds one where no page exists with that name already. (Presumably, the vast majority of the time this would be the first random number it generates.)

naresh-kumar-babu:
> Sorry for the delay. One issue here is that the page and file name are too hardcoded - the code should ensure that a page and file with this name don't already exist. One easy way to do that is to add a random number to the end so that, for example, temp_page is not set to "Canasta-Wiki-Check-Temp-Page", but rather "Canasta-Wiki-Check-Temp-Page-XXX", where "XXX" is a random three-digit number. That way the code can cycle through random three-digit numbers until it finds one where no page exists with that name already. (Presumably, the vast majority of the time this would be the first random number it generates.)

Updated.

yaronkoren:
Looks good!

cicalese:
Thanks for this, and for adding unit tests alongside it — `TestWikiCheckWrite` covering the success path is appreciated.

The following all need to be addressed before this can merge.

1. **`--write` should become a subcommand.** This is already agreed on our end, so please go ahead and restructure it — the write checks belong behind their own subcommand rather than a flag on the existing command. Along with that, the command needs to make clear to the user that it modifies the wiki: a warning before it runs and a confirmation prompt, with a `-y`/`--yes` flag to skip the prompt for scripted use.

2. **`_get_admin_username` can't assume user ID 1.** It runs `User::newFromId(1)->getName()` and falls back to `WikiSysop`. That holds for a wiki Canasta created itself, but not for one whose database was imported — a migration from an existing wiki, a restore, or `canasta import`. In those cases user 1 is whoever happened to be created first on the original wiki: possibly a real human contributor, possibly a bot, possibly a renamed or deleted account, possibly someone with no admin rights at all. The user may also simply not exist, in which case the name comes back empty or unusable and we fall through to `WikiSysop`, which on an imported wiki may not exist either.

   The consequence is worse than a failed check: `createBotPassword` would provision standing write, delete, and upload credentials on an arbitrary user's account without their knowledge, persist them to disk, and then attribute the test edits, uploads, and deletions to that user in RecentChanges and the logs.

   Please don't infer the account. Take the username explicitly, or resolve one that actually holds the required rights, and fail with a clear message if neither is available.

3. **Credential written in plaintext where gitops will commit it.** `_get_or_create_bot_password` writes `admin-bot-password_<wiki_id>` into the instance directory. `roles/gitops/files/gitignore.default` ignores `admin-password_*`, which does not match `admin-bot-password_*`, so on a gitops instance the next `canasta gitops push` would commit a live bot password to the repo. Whatever else changes, this needs to not happen — either add a matching ignore rule, or don't persist the credential to disk at all.

4. **TLS verification is disabled on authenticated requests.** `MediaWikiClient` uses `ssl._create_unverified_context()` for all https traffic. That was defensible for the existing unauthenticated reachability probe, but this path sends a bot password and a CSRF token. Certificate verification should be on by default here, with an explicit opt-out flag if we need one for self-signed local setups.

5. **The bot password grant is permanent.** `createBotPassword` is called with `basic,createeditmovepage,delete,uploadfile` and nothing ever revokes it. A single `wiki-check` run leaves standing write and delete credentials on the wiki. Either revoke it at the end of the run, or make the persistence and its lifetime an explicit, documented decision rather than a side effect.

6. **Unbounded retry loop.** The `while True` in `_run_write_check` that searches for an unused page/file name has no iteration cap. It's unlikely to spin in practice, but if all 900 candidates exist — or the query returns something unexpected — it hangs with no output. Please bound it and fail with a clear message.

7. **Silent failure on cleanup.** `delete_page` returns without error on `cantdelete`. That code also covers permission failures, so a run that can create pages but not delete them reports success while leaving the test page and file behind. `cantdelete` should surface; `missingtitle` is the only one that's genuinely benign.

8. **The edits aren't marked as bot edits** — no `bot=1` on the edit calls, and the grant list doesn't include `bot`. On an active wiki these land in RecentChanges as ordinary admin edits.

9. **The probe helpers changed from returning a bool to `(ok, api_url, host_header)`,** which makes them harder to reuse as a plain liveness check — a direction I want to go for `canasta list`. Please keep the reachability result separable from the write-check plumbing rather than entangling them further.

cicalese:
Thank you for your work on this. Based upon the questions that this work raised, we've decided to move forward with an extension that exposes a special page for sysops to use to test the wiki.

Issue #1139, opened 2026-07-14
Parameterize container runtime (docker/podman) across all Compose call sites


cicalese:
## Summary

The container runtime is hardcoded as `docker` / `docker compose` throughout the Compose code paths. This blocks running Canasta under an alternative OCI runtime (rootless Podman via `podman-compose`), which is the goal of #1118 / PR #1122.

PR #1122 parameterizes the runtime (`compose_command` / `inspect_command`) in exactly one place — `roles/orchestrator/tasks/start.yml`. That is not enough to make Podman work, and notably misses the path most `canasta` lifecycle commands actually take. This issue tracks converting **all** call sites to a single source of truth so the runtime can be swapped consistently.

## Why `start.yml` alone is insufficient

`canasta start` / `stop` / `restart` on Compose do **not** run the Ansible `start.yml` in the common case. `direct_commands/lifecycle.py:cmd_start` uses a Python fast-path and only falls back to Ansible for **Kubernetes** or when the instance **has sidecars**:

```python
def cmd_start(args):
    if inst.get("orchestrator") in ("kubernetes", "k8s"): return FALLBACK
    if _helpers._instance_has_sidecars(inst): return FALLBACK
    rc = _helpers._run_compose(inst_id, inst, ["up", "-d"])   # common case
```

`_helpers._run_compose` (and the exec/dump helpers) hardcode `["docker", "compose"]`. So a sidecar-less Compose instance ignores `compose_command` entirely — it starts under Podman only during `create` (which runs Ansible) but then can't be stopped, exec'd, backed up, or upgraded under the same runtime.

### Python fast-path — `direct_commands/_helpers.py`
The primary path for `start`/`stop`/`restart`/exec. All hardcode `docker`/`docker compose`:
- `_run_compose` — lines ~361, ~370 (local + remote/SSH form)
- `_dump_compose_failure` — lines ~560, ~571
- exec helpers — lines ~648, ~658, ~724, ~732
- web-cid lookup — lines ~985, ~993

### Ansible tasks (`roles/**/*.yml`)
`docker` / `docker compose` / `docker inspect` / `docker run` hardcoded in:

| File | Notes |
| --- | --- |
| `roles/orchestrator/tasks/exec.yml` | **central exec abstraction** — used repo-wide |
| `roles/orchestrator/tasks/start.yml` | partially done in #1122 |
| `roles/orchestrator/tasks/stop.yml` | |
[...]

PR #1224, opened 2026-07-27
refactor: parameterize the container runtime as compose_command/inspect_command

cicalese:
Part of #1139 and #1118. Part 2 of 5 splitting #1122 (which stays open as the tracking PR).

**Based on #1223** — review that first; this PR's diff is only meaningful on top of it.

Refactor authored by @hexmode; split out of #1122.

## What

Replaces the hardcoded `docker compose` and `docker` invocations with `compose_command` / `inspect_command` across the Ansible roles and the Python `direct_commands` fast path. Both default to Docker, so **this PR is intended to be a no-op for existing installs**.

Values resolve from the instance's registry entry (new `composeCommand` / `inspectCommand` fields), falling back to the instance `.env`.

This is the piece that closes the gap raised in review on #1122: `canasta start`/`stop`/`restart` for sidecar-less Compose instances run `_helpers._run_compose`, not Ansible, so parameterizing only the Ansible side would have left the runtime override inert on the path most users hit.

## One fix on top of the split

`fix(crowdsec): split compose_command when building the cscli argv` is not @hexmode's work — it is a fix for a regression the split surfaced, committed separately.

`_cscli_argv` is consumed by `command:` with `argv:`, which execs directly without a shell. Building it as `['{{ compose_command }}', 'exec', ...]` yields a single element `"docker compose"`, so every crowdsec subcommand (`ban`, `unban`, `alerts`, `reload`, `console enroll`) would look for a binary of that literal name — on plain Docker, not just Podman. Fixed with `compose_command.split()`, plus a regression test, since the existing test asserted only on the string prefix form and stayed green.

## Design note: the extra_vars shim

`canasta.py` unconditionally injects `compose_command` / `inspect_command` into `extra_vars`. That is not a convenience default — it is load-bearing, and worth understanding before review.

`roles/orchestrator/defaults/main.yml` declares both, but role defaults are scoped to their role, and `canasta.yml` has no `roles:` block — it dispatches per command via `include_tasks` / `include_role`. Three roles reference the variables from outside orchestrator's scope with no `| default(...)`:

| File | Bare references |
|---|---|
| `roles/upgrade/tasks/migrations/mysql_to_mariadb.yml` | 11 |
| `roles/devmode/tasks/enable.yml` | 6 |
| `roles/crowdsec/tasks/_resolve_cscli.yml` | 2 |
| `roles/crowdsec/tasks/_restart_engine.yml` | 1 |

Remove the injection and those 20 references become undefined-variable errors, taking out `canasta crowdsec ban`, `canasta devmode enable`, and the MySQL to MariaDB migration. The injection here is unconditional, so local and `--host` runs are both covered.

**The cost** is that extra-vars sit at the top of Ansible's precedence ladder (role defaults < `include_vars` < `set_fact` < extra-vars), so nothing lower can override them. That is inert in this PR, since the value is always the Docker default, but it constrains what #1226 can do — see the discussion there.

**Suggested follow-up, not folded in here:** `canasta.yml`'s `pre_tasks` already solves this exact problem twice, for `vars/secret_classification.yml` and `vars/tool_versions.yml`, with a comment naming the reason ("roles/*/defaults are role-scoped"). A `vars/compose_runtime.yml` loaded the same way would sit above role defaults but below `set_fact` and extra-vars — resolving all 20 references while still letting a runtime probe or an explicit `-e` win. That would replace the Python shim rather than patch it, and it deserves its own review.

### DIFF
--- a/canasta.py
+++ b/canasta.py
@@ -36,7 +36,7 @@ EXIT_ALREADY_EXISTS = 3
 # so `import canasta_config` resolves the same file Ansible ships.
 sys.path.append(os.path.join(SCRIPT_DIR, "roles", "common", "module_utils"))
 import canasta_config  # noqa: E402
-from direct_commands import _helpers  # noqa: E402
+from direct_commands._helpers import _read_env, _resolve_compose_cmd  # noqa: E402
 
 # Ensure Ansible uses the repo's config regardless of working directory
 os.environ.setdefault("ANSIBLE_CONFIG", ANSIBLE_CFG)
@@ -1045,7 +1045,7 @@ def handle_interactive_exec(args):
         # payload reaches the command: `docker compose exec -T` (no TTY).
         # Without it, omit -T to preserve the interactive shell behavior.
         stdin_file = getattr(args, "stdin_file", None)
-        docker_cmd = _helpers._resolve_compose_cmd(inst) + ["exec"]
+        docker_cmd = _resolve_compose_cmd(inst) + ["exec"]
         if stdin_file:
             docker_cmd.append("-T")
         docker_cmd += [service] + command
@@ -1462,10 +1462,10 @@ def build_ansible_args(ansible_playbook, command_name, args, data):
             ) if os.path.isfile(get_config_file_path()) else {}
             path = inst.get("path", "")
             if path:
-                env_val = _helpers._read_env(path, "compose_command")
+                env_val = _read_env(path, "compose_command")
                 if env_val:
                     extra_vars["compose_command"] = env_val
-                env_val = _helpers._read_env(path, "inspect_command")
+                env_val = _read_env(path, "inspect_command")
                 if env_val:
                     extra_vars["inspect_command"] = env_val
         except SystemExit:
--- a/canasta.py
+++ b/canasta.py
@@ -833,27 +833,6 @@ def get_config_file_path():
     return canasta_config.config_path(get_config_dir())
 
 
-def _read_env_file_value(path, key):
-    """Read a single value from an instance's .env file.
-
-    Parses KEY=VALUE lines. Returns the value string or None.
-    """
-    env_file = os.path.join(path, ".env")
-    try:
-        with open(env_file, "r") as f:
-            for line in f:
-                line = line.strip()
-                if not line or line.startswith("#"):
-                    continue
-                if "=" in line:
-                    k, v = line.split("=", 1)
-                    if k.strip() == key:
-                        return v.strip()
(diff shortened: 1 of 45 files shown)


cicalese:
@hexmode — heads up on two things before this merges.

**`c3cac4e` crashed the CLI, and CI was green on it.** `canasta.py` calls `_helpers._read_env(...)` but never imports `_helpers`, so `build_ansible_args` raised `NameError: name '_helpers' is not defined`. Reproduced by calling the function directly against a temp registry.

Narrowing `except Exception` to `except (OSError, IOError, KeyError)` is what made it fatal rather than invisible — `NameError` isn't in that tuple, so it propagated instead of being swallowed. Both changes are reasonable on their own; together they took out every Ansible-dispatched command against a registered instance.

All seven checks passed on that commit — 1885 unit tests, both CodeQL analyses, Lint, Docker Smoke Test — because nothing in the suite ever called `build_ansible_args`. I've added `tests/unit/test_build_ansible_args_env_fallback.py`, which exercises it against a temporary registry; I verified it fails on `c3cac4e` and passes now. Your narrowed `except` is kept, since it's the right call once the import exists.

**Your force-push dropped `5e5c52f`.** No harm done — I've cherry-picked it back on top of your commit rather than over it, so `c3cac4e` is intact. That commit covers twelve executed call sites the original substitution missed (`canasta exec`, the restic backup invocation, `compose cp` out of a container, the cleanup `compose run`, image prune on purge, and eight image build/tag/push calls), plus a guard test that walks executed values only — Ansible `cmd:`/`argv:`/`*_cmd` and command literals in Python — and fails on any new hardcoded runtime.

I also pulled your `38d73d0` style change from #1225 into this branch so the import style is consistent here, with your authorship. #1225 no longer needs it and I'll drop it there when I rebase the rest of the stack.

**Current state:** 1891 unit tests, lint and validate clean. The full integration suite passed on this branch's content via `workflow_dispatch` — all 12 jobs including Core Lifecycle, GitOps, Kubernetes, Remote Host, and both Features suites. Worth knowing that trigger exists: the PR-event run skips all seven integration jobs, so `gh workflow run tests.yml --ref <branch>` is the only way to get real coverage before merge.

One open item I'd rather flag than quietly leave: I parameterized the image `build`/`tag`/`push` calls that populate the in-cluster registry, but `podman push` to an insecure local registry generally needs `--tls-verify=false`, so swapping the binary alone won't make that path work under Podman. The hardcoding is gone and Docker behavior is unchanged; actual Podman support for it is still outstanding.

Issue #1222, opened 2026-07-27
Container-state queries need a form that works on runtimes without 'compose ps <service>'


cicalese:
## Problem

Several code paths determine container state with a compose subcommand that takes a service name positionally:

- `docker compose ps -q web` — `direct_commands/_helpers.py` (`_check_running_compose`, `_gather_instance_info`), `roles/orchestrator/tasks/check_running.yml`, `roles/orchestrator/tasks/start.yml`
- `docker compose ps --services --filter status=running` — `roles/orchestrator/tasks/list_running_services.yml`, `roles/crowdsec/tasks/_preflight.yml`
- `docker compose ps web --format '{{.Image}}'` — `roles/orchestrator/tasks/upgrade_rebuild_buildable.yml`

`podman-compose ps` does not accept a service name as a positional argument, so none of these have a podman-compose equivalent. They need a form that works across runtimes.

## Constraint: the replacement must stay project-scoped

The obvious substitute is to query the native runtime and select by the label compose sets on the containers it creates:

```
docker ps --filter label=com.docker.compose.service=web --format '{{.ID}}'
```

This is **not** equivalent as written. `docker ps` is a daemon-level query — `chdir` / `cwd` has no effect on it — so this matches the `web` service of *every* compose project on the host, not the instance being asked about. Confirmed by running it from an unrelated directory and getting another instance's container back.

On a host with two or more Canasta instances that means:

- `canasta list` reports every stopped instance as RUNNING as soon as any one instance is up
- `start.yml`'s `_start_web_cid` becomes multi-line and is passed as a single argv element to `inspect -f`, so the health gate either inspects a foreign container or spins the full 60 × 10s before failing
- `canasta upgrade` compares against whichever instance `docker ps` happens to list first
- `list_running_services` and the crowdsec preflight report another instance's services

Multiple instances per host is a supported, documented configuration, so any replacement has to carry a project predicate.

## Correction: the project name is already deterministic

My original write-up framed this as a design decision with a migration problem. That was wrong — there is nothing to decide about deriving the name.

`roles/create/tasks/main.yml:49` builds the instance directory as `{{ _base_path.stdout }}/{{ id }}`, and it is the only place a path is constructed; every other task reads the stored path from the registry. Compose derives the project name from that directory's basename, so **project name == instance id** for every CLI-created instance. The repo already depends on this: `_validate_inputs.yml` filters stale volumes by `name={{ id }}_`.

[...]

PR #1225, opened 2026-07-27
fix: query container state via the runtime's ps with compose labels

cicalese:
Addresses #1222. Part 3 of 5 splitting #1122 (which stays open as the tracking PR).

**Based on #1224** — review that first.

Authored by @hexmode; split out of #1122 **unchanged**, which means the project-scoping problem described in #1222 is still present in this branch. It is isolated here precisely so it can be fixed or reverted without touching the rest of the Podman work.

## What

`podman-compose ps` does not accept a service name as a positional argument, so `compose ps -q web`, `compose ps --services --filter status=running`, and `compose ps web --format` have no podman-compose equivalent. This queries the native runtime instead and selects by the `com.docker.compose.service` label.

## Blocking before merge: this is not project-scoped

`docker ps` is a daemon-level query — `chdir` / `cwd` does not scope it — so `--filter label=com.docker.compose.service=web` matches that service in **every** compose project on the host. Verified by running it from an unrelated directory and getting another instance's container back.

On a host with two or more instances:

- `canasta list` reports stopped instances as RUNNING once any one instance is up
- `start.yml`'s `_start_web_cid` goes multi-line and is passed as a single argv element to `inspect -f`, so the health gate inspects a foreign container or spins the full 60 × 10s
- `canasta upgrade` compares against whichever instance `docker ps` lists first
- `list_running_services` and the crowdsec preflight report another instance's services

Options are written up in #1222. The cheapest is to keep `compose ps` on the Docker path and use the label form only for Podman — but that still needs a project predicate on the Podman path.

## Also needs confirming

`{{.Label "com.docker.compose.service"}}` is used in `list_running_services.yml` and `crowdsec/_preflight.yml`. Podman's `ps` template context exposes `Labels` as a map; which Podman version was the Docker-compat `.Label` *function* validated against?

### DIFF
--- a/direct_commands/_helpers.py
+++ b/direct_commands/_helpers.py
@@ -969,6 +969,19 @@ def _unquote(value):
     return value
 
 
+def _compose_project(path):
+    """The compose project name for an instance directory.
+
+    Compose derives it from the directory basename, lowercased — and
+    rejects uppercase outright, so the lowering is not optional. The
+    CLI creates instance directories named after the instance id, whose
+    other permitted characters are already valid to compose.
+    """
+    if not path:
+        return ""
+    return os.path.basename(os.path.abspath(path)).lower()
+
+
 def _resolve_compose_cmd(inst):
     """Resolve the compose command for an instance.
 
@@ -1067,7 +1080,17 @@ def _check_running(instance_id, path, orchestrator, host, docker_host=None,
 def _check_running_compose(path, host, docker_host=None, compose_cmd=None):
     if compose_cmd is None:
         compose_cmd = ["docker", "compose"]
-    ps_cmd = compose_cmd + ["ps", "-q", "web"]
+    # podman-compose's `ps` takes no service argument, so query the
+    # native runtime by label instead. `ps` is a daemon-level call —
+    # cwd does not scope it — so the project filter is what keeps this
+    # from matching another instance's web container.
+    runtime = "podman" if "podman" in compose_cmd[0] else "docker"
+    ps_cmd = [
+        runtime, "ps",
+        "--filter", "label=com.docker.compose.project=%s" % _compose_project(path),
+        "--filter", "label=com.docker.compose.service=web",
+        "--format", "{{.ID}}",
+    ]
     if _is_localhost(host):
         try:
             result = subprocess.run(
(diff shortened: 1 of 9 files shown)


cicalese:
@hexmode — #1224 just merged, so this branch needs a rebase onto main, and that means a force-push. Could you hold off pushing here for a few minutes and let me know when you're at a stopping point? I'll do it and confirm as soon as it lands.

Why it needs rewriting rather than a merge: #1224 was squash-merged, so its commits never became ancestors of main. This branch still carries `25d3e32`, `5e5c52f` and your `38d73d0`, all of which are now in main via that squash — `38d73d0` came along because I cherry-picked it into #1224 so the import style was consistent there. Rebasing drops all three as already-applied and replays only `7f32f24`, the container-state query change, which is what this PR is actually about. The diff should shrink to roughly six files.

If you have unpushed work, push it first and I'll rebase on top of it — I'd rather not rewrite out from under you. Otherwise just say go.

Two heads-ups for when it does land:

The unscoped-label problem in #1222 is still open here, deliberately. `docker ps --filter label=com.docker.compose.service=web` is a daemon-level query, so `chdir` doesn't scope it and it matches that service in every compose project on the host. On a host with two instances, `canasta list` reports stopped ones as RUNNING and `start.yml`'s health gate can inspect a foreign container. It needs a project predicate; the options are written up in #1222 and the choice is yours.

Also worth knowing: `gh workflow run tests.yml --ref split/3-container-state-queries` runs the full suite including the seven integration jobs, which the PR-event trigger skips. Given this PR changes how running-state is determined, that's worth doing before merge — the PR checks alone won't exercise it.

cicalese:
@hexmode — pushed. Thanks for the go-ahead.

Heads up that I found you'd rebased all three yourself just before I pushed, so I checked your versions rather than assuming mine won. Two reasons mine superseded, and **nothing of yours was dropped**:

- Your rebases landed on the pre-merge #1224 branch tip (`73ea47a`) rather than on merged main. Since #1224 was squash-merged, that left its commits still in the history here, so this PR's diff included them.
- Your `split/4` and `split/5` tips failed two tests. `test_build_ansible_args_env_fallback` asserts the blanket `extra_vars` default, which the play-scope precedence fix in #1226 deliberately removes — so that test has to be updated inside that commit. It now asserts the *absence* of the extra-var, which documents the contract change rather than papering over it.

The only other difference was the `canasta.py` import style: your branches still had `from direct_commands import _helpers`, because they predate the cherry-pick of your `38d73d0` into #1224. Main now carries your direct-import form, so that's what these are on — your style, just arriving via main.

Where things stand:

| PR | base | files | state |
|---|---|---|---|
| #1225 | `main` | 7 | CI running |
| #1226 | `split/3` | 10 | CLEAN |
| #1227 | `split/4` | 15 | CLEAN |

#1225 is down to the 7 files it's actually about. All three pass locally — 1891 / 1906 / 1906, lint and validate clean.

#1225 is the only one whose base is `main`, so it's the only one CI will run on. For real coverage: `gh workflow run tests.yml --ref split/3-container-state-queries`.

Still open here and unchanged: the label filter isn't project-scoped (#1222). On a multi-instance host `canasta list` will report stopped instances as RUNNING. That one needs a decision from you on how the compose project name is derived.

PR #1233, opened 2026-07-28
fix(crowdsec): pass the service-label query as argv, not through a shell

### LINKED ISSUE #1231: All crowdsec Compose subcommands fail: unquoted --format in docker ps splits under the shell
Found in a hands-on Compose e2e against merged `main` (`339da8f`). Regression from #1225.

## Symptom

Every `canasta crowdsec *` subcommand fails on Compose, even with the crowdsec service running:

```
$ docker ps --filter label=com.docker.compose.project=e2e-compose \
            --filter label=com.docker.compose.service=crowdsec --format '{{.Status}}'
Up About a minute

$ canasta crowdsec status -i e2e-compose
Error: The command exited with a non-zero return code.
 (cmd: docker ps --filter status=running --filter label=com.docker.compose.project=e2e-compose --format {{.Label "com.docker.compose.service"}})
docker: 'docker ps' accepts no arguments

Usage:  docker ps [OPTIONS]
```

## Cause

`roles/crowdsec/tasks/_preflight.yml:26` and `roles/orchestrator/tasks/list_running_services.yml:13` both build the command with an **unquoted** `--format` value inside an `ansible.builtin.shell:` task:

```yaml
cmd: >-
  {{ inspect_command | default('docker') }}
  ps --filter status=running
  --filter label=com.docker.compose.project={{ instance_path | basename | lower }}
  --format {% raw %}{{.Label "com.docker.compose.service"}}{% endraw %}
```

`shell:` runs it through `/bin/sh`, which word-splits the format on the space inside it, so docker receives `{{.Label` and `com.docker.compose.service}}` as two extra positional arguments and rejects the whole invocation:

```
$ sh -c 'docker ps --filter status=running --format {{.Label "com.docker.compose.service"}}'
docker: 'docker ps' accepts no arguments

$ sh -c 'docker ps --filter status=running --format "{{.Label \"com.docker.compose.service\"}}"'
varnish
web
crowdsec
caddy
db
```

The predecessor, `docker compose ps --services --filter status=running`, had no format string and so no quoting hazard.

## Impact

- **`roles/crowdsec/tasks/_preflight.yml`** — the task fails outright, taking down `crowdsec status`, `ban`, `unban`, `alerts`, `metrics`, `scenarios`, `reload` and `console enroll` on Compose. Note the guard below it (`when: "'crowdsec' not in (_crowdsec_running.stdout_lines | default([]))"`) would *also* misfire on an empty result, refusing with "The crowdsec service is not running" while it plainly is.
- **`roles/orchestrator/tasks/list_running_services.yml`** — consumed by `roles/maintenance/tasks/exec.yml`, which reports "No running services found." regardless of actual state.

Kubernetes is unaffected — that branch uses `kubectl` with a single-quoted jsonpath.

## Fix

Quote the format value, or switch these two tasks from `shell:` to `command:` with an `argv:` list so no shell parsing happens at all. `argv:` is the more robust option and matches how `start.yml` passes its `-f` template to `docker inspect`.

## Coverage gap

The integration suite has an `Integration: Kubernetes CrowdSec` job but no Compose equivalent, which is why this shipped green. A Compose crowdsec scenario — enable, `status`, `ban`, `unban` — would have caught it, and would be cheap to add alongside the existing Features jobs.

cicalese:
Fixes #1231. Found in a hands-on Compose e2e against merged `main`.

## The bug

`--format {{.Label "com.docker.compose.service"}}` sat unquoted inside an `ansible.builtin.shell:` task. The template contains a space, so `/bin/sh` split it into `{{.Label` and `com.docker.compose.service}}`, and docker rejected the whole invocation:

```
$ canasta crowdsec status -i e2e-compose
Error: The command exited with a non-zero return code.
 (cmd: docker ps --filter status=running --filter label=com.docker.compose.project=e2e-compose --format {{.Label "com.docker.compose.service"}})
docker: 'docker ps' accepts no arguments
```

That broke every crowdsec Compose subcommand, and `canasta backup restore`, which reaches the same query through the start path's CrowdSec autoenroll — aborting at exit 2 before restoring anything.

## The fix

Both tasks move from `shell:` to `command:` with `argv:`. Each element is one argument regardless of its contents, so no shell parsing happens and the template arrives intact. This matches how `start.yml` already passes its `-f` template to `docker inspect`.

Verified against the live instance the bug was found on — the error is gone and the query returns the service list correctly:

```
varnish
web
caddy
db
```

## Test

`test_container_query_no_shell_splitting.py` checks that no Go template containing whitespace is left unquoted, across all five files that query the runtime by template. It only inspects text inside `{% raw %}` — Ansible resolves its own `{{ jinja }}` before the argument vector is built, so those are not a hazard, and an earlier draft of this test produced a false positive on `{{ inspect_command | default('docker') }}` for exactly that reason.

Confirmed it fails on unmodified `main` and passes with the fix.

## Note on the guard below it

`_preflight.yml`'s `when: "'crowdsec' not in (_crowdsec_running.stdout_lines | default([]))"` would also have misfired on the empty result, refusing with "the crowdsec service is not running" while it was running. Fixing the query fixes that too, but the guard is worth a look independently — `default([])` turns a failed probe into a confident negative.

## Coverage gap

The suite has an `Integration: Kubernetes CrowdSec` job and no Compose equivalent, which is why this shipped green through five PRs. A Compose scenario — enable, `status`, `ban`, `unban` — would have caught it and is cheap to add.

### DIFF
--- a/roles/crowdsec/tasks/_preflight.yml
+++ b/roles/crowdsec/tasks/_preflight.yml
@@ -18,12 +18,16 @@
   when: instance_orchestrator | default('compose') == 'compose'
   block:
     - name: List running services
-      ansible.builtin.shell:
-        cmd: >-
-          {{ inspect_command | default('docker') }}
-          ps --filter status=running
-          --filter label=com.docker.compose.project={{ instance_path | basename | lower }}
-          --format {% raw %}{{.Label "com.docker.compose.service"}}{% endraw %}
+      ansible.builtin.command:
+        argv:
+          - "{{ inspect_command | default('docker') }}"
+          - ps
+          - --filter
+          - status=running
+          - --filter
+          - "label=com.docker.compose.project={{ instance_path | basename | lower }}"
+          - --format
+          - '{% raw %}{{.Label "com.docker.compose.service"}}{% endraw %}'
         chdir: "{{ instance_path }}"
       register: _crowdsec_running
       changed_when: false
--- a/roles/orchestrator/tasks/list_running_services.yml
+++ b/roles/orchestrator/tasks/list_running_services.yml
@@ -5,12 +5,16 @@
 # Output: _running_services (newline-separated names, may be empty)
 
 - name: Get service list (Compose)
-  ansible.builtin.shell:
-    cmd: >-
-      {{ inspect_command | default('docker') }}
-      ps --filter status=running
-      --filter label=com.docker.compose.project={{ instance_path | basename | lower }}
-      --format {% raw %}{{.Label "com.docker.compose.service"}}{% endraw %}
+  ansible.builtin.command:
+    argv:
+      - "{{ inspect_command | default('docker') }}"
+      - ps
+      - --filter
+      - status=running
+      - --filter
+      - "label=com.docker.compose.project={{ instance_path | basename | lower }}"
+      - --format
+      - '{% raw %}{{.Label "com.docker.compose.service"}}{% endraw %}'
     chdir: "{{ instance_path }}"
   register: _list_services_compose
   changed_when: false
--- a/tests/unit/test_container_query_no_shell_splitting.py
+++ b/tests/unit/test_container_query_no_shell_splitting.py
@@ -1,0 +1,107 @@
+"""A Go template containing whitespace must never be passed unquoted.
+
+`{{.Label "com.docker.compose.service"}}` has a space in it. Handed to
+`ansible.builtin.shell:` without quoting, /bin/sh splits it into two
+words and docker rejects the call with "docker ps accepts no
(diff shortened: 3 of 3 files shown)


cicalese:
Correction to the last section of the description: the note about `default([])` turning a failed probe into a confident negative is wrong, and I should not have raised it.

The probe task carries neither `failed_when: false` nor `ignore_errors`, so a failure aborts the play — which is exactly what we saw in #1231, where the run stopped at the `docker ps` error rather than proceeding to a bogus refusal. The `default([])` is only reached when the task is *skipped*, and refusing in that case fails closed, which is the right behavior. Nothing to change there.

The fix itself and the test are unaffected.

Issue #1282, opened 2026-07-29
create's rescue cleanup cannot remove the instance directory under rootless Podman (EACCES on config/)


cicalese:
When `canasta create` fails and its rescue path runs, the "Clean up on failure (remove directory)" task in `playbooks/create.yml` cannot delete the instance directory under rootless Podman:

```
[ERROR]: Task failed: Module failed: rmtree failed: [Errno 13] Permission denied:
'/tmp/canasta-int-work-gij7xuuf/inttest-lifecycle/config'
Origin: playbooks/create.yml:63:7
```

Under rootless Podman the containers write through a user-namespace ID map, so files the stack creates in bind-mounted directories end up owned by a mapped subuid rather than by the invoking user. The controller user cannot then remove them, so the cleanup that is supposed to leave no trace instead leaves a directory the operator cannot delete either.

## Why it matters

The rescue exists so a failed create does not strand state. When it fails, the operator is left with an instance directory that blocks a retry — and the next `create` refuses to enter a non-empty target directory. That is the same failure shape as #1267, where `delete` left an undeletable instance directory on Podman; this is the `create`-rescue counterpart.

## Reproduction

Any `create` failure on rootless Podman reaches it. It surfaced behind #1281 (the web health gate failing on Podman 4.9.3), so #1281 has to be fixed first for this to be reachable in a normal run.

Environment: Podman 4.9.3, podman-compose 1.3.0, Ubuntu 24.04, rootless.

## Notes toward a fix

Whatever #1267 settled on for `delete` is the natural reference point — the two paths should tear down the same way rather than each solving ID-mapped ownership separately.

## How it was found

The Podman CI lane added for #1279.

Issue #1279, opened 2026-07-28
CI has no Podman coverage: the integration suite passed green while upgrade was broken on Podman


cicalese:
Podman is now a supported container runtime, and no CI job exercises it. `grep -rn podman .github/workflows/` returns nothing.

## This is demonstrable, not theoretical

At commit `0a2550c`, `canasta upgrade` could not pull a single image on a Podman host — it died on `podman-compose: error: unrecognized arguments: --ignore-buildable --ignore-pull-failures` (#1275).

The post-merge integration run for that same commit:

```
0a2550c Tests → success
```

That run includes the `upgrade` leg. It passed because it ran against Docker, which is the only runtime CI has ever used.

## How this played out

Five Podman defects were found by a person running the CLI by hand, not by CI:

| | |
|---|---|
| #1264 | `COMPOSE_PROFILES` read on the controller — profiled services, database included, silently dropped from `up` |
| #1265 | `create` reported success with nothing running |
| #1267 | `delete` left an undeletable instance directory |
| #1270 | `upgrade` ignored the instance's recorded runtime |
| #1275, #1276 | Docker-only flags handed to podman-compose |

They surfaced one at a time, each only after the previous fix let the run advance one step further — three separate merge cycles, each landing on `main` before the next problem was visible.

#1276 is the one that should worry us most: `upgrade_rebuild_buildable.yml` carries `failed_when: false`, so `config --format json` failing on podman produced an empty service list and buildable services were **silently never rebuilt**, while the upgrade reported success. No amount of hand-testing reliably catches a silent skip. Only an assertion does.

## Two distinct gaps

**1. No Podman runtime in CI.** The more important one. Even a post-merge-only Podman lane would have caught all five.

**2. Integration jobs are gated `if: github.event_name != 'pull_request'`** (see #67), so a PR's green checks say nothing about integration. Combined with gap 1, a Podman-breaking change gets two green signals — PR checks and the post-merge run — and still ships broken. This gating is presumably deliberate (cost, secrets), so it is context here rather than a request to change it.

## Suggested scope

A `integration-podman` job mirroring `integration-core`'s `lifecycle upgrade` legs with podman + podman-compose installed and Docker absent, gated the same way as the existing integration jobs. That is the minimum that would have failed on #1275 and #1276.

[...]

Issue #67, opened 2026-04-12
CI: only run integration test jobs on pushes to main


cicalese:
## Background

Today's CI runs every job (Unit Tests, Lint, Helm Chart Lint, Docker Smoke Test, and **five integration jobs**) on every PR. The five integration jobs each spin up a real Canasta instance (or a real k3s cluster, or a real remote host), which makes them genuinely slow:

- `Integration: Core Lifecycle`
- `Integration: Features`
- `Integration: GitOps`
- `Integration: Kubernetes` (installs k3s in CI)
- `Remote Host Integration Tests`

A typical PR run takes 15-20 minutes wall-clock waiting for the integration tail to finish, even when the diff is one-file documentation or a one-line threshold tweak that couldn't possibly affect integration behavior.

This week's batch of small follow-up PRs (#56 → #60 → #61 → #62 → #64 → #65 → #66) made the cost concrete: we're spending more wall-clock waiting on integration tests than reviewing the actual code.

## Proposal

**Run the five integration jobs only on pushes to `main` and manual `workflow_dispatch` triggers.** Keep the fast jobs (Unit Tests, Lint, Helm Chart Lint, Docker Smoke Test) running on every PR.

PR feedback drops from ~15-20 minutes to ~2 minutes for the common case, while integration coverage on `main` is unchanged — anything that lands gets the full suite immediately on the merge commit. If a regression sneaks past the fast jobs and lands on main, the next merge will catch it (fail-loudly on main) and the fix is a follow-up PR. That trade is acceptable given how often we've been doing follow-up PRs anyway.

The `workflow_dispatch` trigger gives us a manual escape hatch: for PRs that touch integration-relevant code (the chart, the create/start/stop tasks, the k8s preflight, etc.), the reviewer can click "Run workflow" on the Actions tab to get a full integration run on demand.

## Concrete change

`.github/workflows/tests.yml`:

1. Add `workflow_dispatch:` to the `on:` block.
2. Add `if: github.event_name != 'pull_request'` to each of the five integration jobs.

That's it. No new files, no test changes, no logic changes — just gating.

## What still runs on PRs

- `unit` (Unit Tests + coverage) — fast, catches the most regressions per minute
- `lint` (yamllint + validate_definitions.py) — fast, catches structural problems
- `helm-lint` (Helm Chart Lint + template render) — fast, catches chart bugs
[...]

PR #1283, opened 2026-07-29
fix(start): skip the health gate when the runtime reports no health status

### LINKED ISSUE #1281: canasta create fails on Podman 4.9.3: the web health gate waits 10 minutes on a status that never appears
On Podman 4.9.3 with podman-compose 1.3.0 (Ubuntu 24.04), `canasta create` never completes. The web health gate in `roles/orchestrator/tasks/start.yml` retries 60 × 10s and then fails the play:

```
fatal: [localhost]: FAILED! => {"attempts": 60, "changed": false,
  "cmd": ["podman", "inspect", "-f", "{{.State.Health.Status}}", "301218948b65"],
  "rc": 0, "stderr": "", "stdout": ""}
```

`rc: 0`, empty stdout, for the full ten minutes.

## Why the existing guard does not catch it

`start.yml` already anticipates a runtime with no health data. The task at line 169 probes

```
{{if .State.Health}}has{{end}}
```

and the wait at line 184 is gated on that rendering `has`. Since the wait ran to exhaustion, the guard must have matched — so on Podman 4.9.3 `.State.Health` is present as a non-nil struct while `.State.Health.Status` renders empty. Testing for the struct's presence is therefore not a usable proxy for "this container reports health".

The comment above the guard records where it *was* verified:

> Podman note: `podman inspect -f '{{.State.Health.Status}}'` returns the same Go template output as Docker for containers with a HEALTHCHECK. Validated with Podman 4.3.1 (Debian 12 Bookworm ARM64). Earlier rootless Podman (<4.0) may not populate .State.Health at all; the `has` probe below skips gracefully in that case.

So the behavior differs between Podman 4.3.1 and 4.9.3, and the `has` probe does not in fact skip gracefully on the newer one.

## Impact

`create` is unusable on this Podman/OS combination — it fails after a ten-minute stall, and the failure is a bare empty-string comparison that gives the operator nothing to go on. `lifecycle`, `upgrade`, and `upgrade-rebuilds-buildable` all fail this way; `doctor` passes, so runtime detection itself is fine.

## Suggested fix

Gate on the rendered status being non-empty rather than on the struct being non-nil — i.e. probe `{{if .State.Health}}{{.State.Health.Status}}{{end}}` and skip the wait when that comes back empty. An absent or unpopulated field then reads as "no health signal to wait on" on every runtime and version, with no per-version knowledge encoded.

Worth deciding separately whether skipping the wait is safe enough on Podman: the comment above the gate explains it exists to stop `up -d` racing composer. Skipping restores that race on runtimes with no health signal, which is what already happens on Podman < 4.0 today.

## How it was found

The Podman CI lane added for #1279, on its first successful run. This is exactly the version-dependent difference that lane exists to surface — it is not reachable by hand-testing a single Podman version.

cicalese:
Fixes #1281.

`canasta create` could not complete on Podman 4.9.3 / podman-compose 1.3.0 (Ubuntu 24.04). The web health gate waited out its full 60 × 10s and then failed the play:

```
fatal: [localhost]: FAILED! => {"attempts": 60,
  "cmd": ["podman", "inspect", "-f", "{{.State.Health.Status}}", "301218948b65"],
  "rc": 0, "stdout": ""}
```

## Cause

The gate was guarded by a probe rendering `{{if .State.Health}}has{{end}}`, on the assumption that a runtime with no health data omits `.State.Health`. Podman 4.9.3 instead reports the struct as non-nil with an empty `Status`, so the guard matched, the wait ran, and the status it was waiting for never appeared. The result was a ten-minute stall ending in an empty-string comparison that gave the operator nothing to work with.

The comment above the guard recorded that it had been verified on Podman 4.3.1 — the behavior differs on 4.9.3.

## Fix

Probe the status itself rather than the presence of the struct, and gate the wait on that render being non-empty. Every "no usable health signal" case now takes the same path — older Podman omitting `.State.Health`, Podman 4.9.3 leaving `Status` empty, and a custom image with no `HEALTHCHECK` — with no per-version knowledge encoded in the playbook.

The `if` guard is kept so `inspect` does not error on runtimes that omit the field entirely.

## Verification

Confirmed against a real Docker daemon that behavior is unchanged there:

| container | old probe | new probe |
|---|---|---|
| with `--health-cmd` | `has` | `healthy` |
| no healthcheck | *(empty)* | *(empty)* |

So a healthy container still gates the wait, and one without a healthcheck still skips it.

`tests/unit/test_start_health_gate_probe.py` pins the behavior: the probe must render the status, keep the `if` guard, and the wait must be gated on a non-empty status rather than a sentinel. 2079 unit tests, yamllint, and `validate_definitions.py` pass.

## Note

This was found by the Podman CI lane in #1279, which is blocked on this fix — the lane cannot go green until `create` works on Podman. #1282 (rescue cleanup hitting EACCES under rootless Podman) sits behind this one and is only reachable once this is in.

Not verified on Podman directly — there is no Podman host in this environment. The #1279 lane will exercise it.

### DIFF
--- a/roles/orchestrator/tasks/start.yml
+++ b/roles/orchestrator/tasks/start.yml
@@ -133,12 +133,12 @@
     # HEALTHCHECK skip the wait (we trust them to define their own
     # readiness contract).
     #
-    # Podman note: `podman inspect -f '{{.State.Health.Status}}'`
-    # returns the same Go template output as Docker for containers
-    # with a HEALTHCHECK.  Validated with Podman 4.3.1 (Debian 12
-    # Bookworm ARM64).  Earlier rootless Podman (<4.0) may not
-    # populate .State.Health at all; the `has` probe below skips
-    # gracefully in that case.
+    # Runtimes disagree on what "no health data" looks like: Podman
+    # 4.9.3 reports a non-nil .State.Health whose Status is empty, older
+    # Podman omits .State.Health entirely, and Docker reports a real
+    # status. So the probe renders the status itself rather than testing
+    # the struct for presence — an empty render means "nothing to wait
+    # on" on every runtime, with no version knowledge encoded.
     - name: Get web container id
       ansible.builtin.command:
         cmd: >-
@@ -166,13 +166,13 @@
           containers as soon as the session that started them ends.
       when: _start_web_cid.stdout | trim == ''
 
-    - name: Probe whether web declares a healthcheck
+    - name: Probe whether web reports a health status
       ansible.builtin.command:
         argv:
           - "{{ _inspect_cmd }}"
           - inspect
           - "-f"
-          - "{% raw %}{{if .State.Health}}has{{end}}{% endraw %}"
+          - "{% raw %}{{if .State.Health}}{{.State.Health.Status}}{{end}}{% endraw %}"
           - "{{ _start_web_cid.stdout | trim }}"
       register: _start_web_has_health
       changed_when: false
@@ -196,7 +196,7 @@
       until: _start_web_health.stdout | trim == 'healthy'
       when:
         - _start_web_cid.stdout | trim != ''
-        - _start_web_has_health.stdout | default('') | trim == 'has'
+        - _start_web_has_health.stdout | default('') | trim != ''
 
     # Register CAPI + auto-enroll the bouncer when CrowdSec is on. Shared
     # with the K8s start path and create. crowdsec_autoenroll is false only
--- a/tests/unit/test_start_health_gate_probe.py
+++ b/tests/unit/test_start_health_gate_probe.py
@@ -1,0 +1,98 @@
+"""The health gate must skip when the runtime reports no health status.
+
+Runtimes disagree on what a container without usable health data looks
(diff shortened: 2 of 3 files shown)


cicalese:
Integration validation: dispatched the full suite against this branch, since integration jobs are gated off on pull requests (#67) and the PR's green checks never start a container.

[Run 30410922229](https://github.com/CanastaWiki/Canasta-CLI/actions/runs/30410922229) — **12/12 success**, including every job that creates and starts a real instance through the rewritten gate:

| Job | Result |
|---|---|
| Integration: Core Lifecycle | success |
| Integration: Features (config & farms) | success |
| Integration: Features (backup & ops) | success |
| Integration: GitOps | success |
| Integration: Kubernetes | success |
| Integration: Kubernetes Backup | success |
| Integration: Kubernetes CrowdSec | success |
| Remote Host Integration Tests | success |
| Unit Tests / Lint / Helm Lint / Docker Smoke | success |

So the probe rewrite does not regress Docker: `create`, `start`, `stop`, `restart`, and `upgrade` all pass through the gate unchanged.

Still unverified on Podman — the lane that exercises it is in #1279, which is blocked on this merging.

Issue #1276, opened 2026-07-28
podman: 'config --format json' unsupported — breaks backup build discovery, silently skips rebuilds


cicalese:
Two tasks render the merged Compose model with `config --format json`. podman-compose 1.3.0 has no `--format` option at all:

```
usage: podman-compose config [-h] [--no-normalize] [--services]
```

```
$ podman-compose config --format json ; echo $?
2
```

## Two call sites, two different symptoms

**`roles/orchestrator/tasks/backup_discover_build_paths.yml:76`** — registers the output and pipes it through `from_json`. No `failed_when`, so this is a hard failure: `canasta backup` cannot discover build paths on a Podman instance.

**`roles/orchestrator/tasks/upgrade_rebuild_buildable.yml:55`** — has `failed_when: false`, so it degrades silently instead. The command fails, no build sections are found, and buildable services are quietly not rebuilt during upgrade. This is the worse of the two: it reports success while skipping work.

## Available equivalent

podman-compose's default `config` output is valid YAML and parses cleanly:

```
$ podman-compose config | python3 -c 'import sys,yaml; print(list(yaml.safe_load(sys.stdin)["services"]))'
['caddy', 'web']
```

So the podman path can drop `--format json` and use `from_yaml` in place of `from_json`.

One behavioral difference worth checking during review: podman-compose does **not** interpolate `${VAR}` in the rendered output, where `docker compose config` does. The backup task carries `no_log: true` specifically because the Docker output embeds MYSQL_PASSWORD; on Podman that exposure does not occur. Since both call sites only read `build` sections, non-interpolation should be harmless — but it should be confirmed rather than assumed.

Both sites already have `when: compose_command is search('podman')` blocks for profile flags, so the runtime distinction is established there; only the `config` invocation was missed.

Verified on `main` at 0a2550c, podman 5.4.2 / podman-compose 1.3.0.

PR #1285, opened 2026-07-29
fix: raise unprivileged port floor to 80 for rootless runtimes, add fix-sysctl command

### LINKED ISSUE #1284: Raise unprivileged port floor to 80 for rootless runtimes at create time, add fix-sysctl command
From PR #1122 review by @cicalese (point 3 in https://github.com/CanastaWiki/Canasta-CLI/pull/1122#pullrequestreview-4787405848):

The original PR included a block in `create_preflight.yml` that silently sudo-edited the host sysctl config via `ansible.posix.sysctl` with `sysctl_set: yes` — persisting the change to `/etc/sysctl.conf` without the operator opting in, and with no `ansible_system == "Linux"` guard.

This issue tracks implementing the fix properly:

1. **Runtime fix in create_preflight.yml**: When rootless Podman is detected and `ip_unprivileged_port_start > 80`, attempt `sysctl -w net.ipv4.ip_unprivileged_port_start=80` as a temporary runtime fix (no persistence). If sudo is unavailable or it fails, fall through to the existing fail message. The change must be gated behind `ansible_system == "Linux"` and live inside the `orchestrator == "compose"` block.

2. **`canasta fix-sysctl` command**: A new direct-only command (like `doctor`) that applies the change *permanently* — runs `sudo sysctl -w` (runtime) and writes `net.ipv4.ip_unprivileged_port_start=80` to `/etc/sysctl.d/canasta-privport.conf` (persistent). Supports `--host` for remote targets and `--id` to resolve the host from a registered instance. Idempotent — if the floor is already ≤ 80, reports nothing to fix.

hexmode:
## Summary

Fixes #1284. Addresses point 3 from the #1122 review — the original PR used `ansible.posix.sysctl` with `sysctl_set: yes` which silently persisted the change to `/etc/sysctl.conf` without opt-in. This implements the proper two-tier approach.

### `roles/orchestrator/tasks/create_preflight.yml`
- Before failing when `ip_unprivileged_port_start > 80` for rootless Podman, attempt a **runtime-only** `sysctl -w net.ipv4.ip_unprivileged_port_start=80` (no persistence).
- Gated behind `ansible_system == "Linux"` and inside the existing `orchestrator == "compose"` block.
- If sudo fails or is unavailable, the existing hard-fail with manual remediation instructions runs.
- On success, a warning is printed reminding the user to run `canasta fix-sysctl` to make it permanent.

### `direct_commands/fix_sysctl.py` — new `canasta fix-sysctl` command
- Direct-only command (like `doctor`).
- Reads the current `ip_unprivileged_port_start` value.
- If already ≤ 80, reports nothing to fix.
- If > 80: applies `sudo sysctl -w net.ipv4.ip_unprivileged_port_start=80` (immediate) and writes to `/etc/sysctl.d/canasta-privport.conf` (persistent).
- Supports `--host` (remote targets) and `--id` (resolve host from instance).
- Reports clear errors if sudo is unavailable.

### `meta/command_definitions.yml` / `direct_commands/__init__.py`
- Command definition and module registration for `fix-sysctl`.

## Verification
- `make lint` passes
- `make validate` passes (88 commands, 69 playbooks)
- `make test-unit` — all 344 tests pass


cicalese:
Thanks for reworking this — much tighter than the original block. Three of the four objections from the split are gone: runtime-only rather than persisted, guarded on `ansible_system == 'Linux'`, and inside the Compose-only block. It's also properly additive; the existing check still runs.

The `create_preflight.yml` half can't land, though. **This project does not escalate privileges outside provisioning code.** On `main` (`e85f48f`), `become:` appears in seven files and every one is provisioning — `install/tasks/{canasta,docker,git_crypt,sops_age}.yml`, `k8s_install_k3s.yml`, `uninstall_k3s.yml`, `storage_setup_nfs.yml`. No instance operation escalates, and every `sudo` in the Python is advice text, not an invocation. Adding `become: yes` to the create path would be the first exception.

We've already settled this exact case once. `gitops init --encrypt-secrets` needs sops, which needs root to install. It refuses rather than escalating (`_init_sops.yml:47`): *"Install the toolchain with `canasta install sops` ... The gitops target host also needs sops: `canasta install -H <host> sops`."* Auto-running `sysctl -w` during create is the same move we declined there.

Two practical problems on top of the principle:

**It defers the failure instead of fixing it.** `sysctl -w` doesn't survive a reboot. Today the operator is stopped at create time and fixes it properly. With this, create succeeds and the problem resurfaces after a reboot as Caddy failing with `rootlessport cannot expose privileged port 80` — with no create running and the reminder long gone from the scrollback.

**It only helps operators who were already unblocked.** `become: yes` needs passwordless sudo; without it the task fails, `failed_when: false` swallows it, and the existing hard-fail runs — identical to today. And that message already gives a no-root remedy: *"or set HTTP_PORT and HTTPS_PORT above 1024 for this instance."*

`fix-sysctl` itself is fine — explicit, operator-invoked escalation is exactly where it belongs. Consider whether it wants to be a new top-level command: `doctor` already detects this condition and prints the same advice (`doctor.py:120-123`), and `canasta install` already takes `docker`/`git-crypt`/`sops`.

---

**`fix_sysctl.py`** — `PERSIST_FAILED` prints a good warning but returns `0`. Persistence is the command's whole purpose, so a scripted caller shouldn't see success. `return 1` with the same message.

**No tests.** This adds a command that shells out to `sudo` and changes when an existing hard-fail fires. Every podman fix in 4.14.0 (#1266, #1268, #1269, #1272, #1277, #1278) shipped with tests asserting the runtime-conditional behavior.

**"all 344 tests pass"** — `main` has 2067. The branch merges cleanly against current `main`, so that reads as a subset run. Please run the full `make test-unit`.

**Nits:** repo uses `become: true`, not `become: yes`; and `sysctl -w ...` wants `ansible.builtin.command`, not `shell`.

cicalese:
Following up on the naming question I raised above — I think the better home is clearer than I put it, and it's `canasta install podman`. Filed as #1288.

`install docker` is the exact parallel. It doesn't just drop a binary (`roles/install/tasks/docker.yml`): it adds the repo, installs, starts the service, and adds the operator to the `docker` group (line 158) — the prerequisite that makes Docker usable without root afterward. Podman's equivalents are the same shape: podman + podman-compose, `loginctl enable-linger`, and the port floor.

Checking this turned up a gap neither of us named: **there is no podman install path in the CLI at all.** `grep -rn podman roles/install/ playbooks/install.yml` returns nothing. Canasta has supported podman as a runtime since 4.14.0, but the operator still provisions the host by hand. That asymmetry is the real problem, and the sysctl floor is one symptom of it — lingering is the other, and it is currently a second manual step in the same preflight failure.

So the concrete suggestion is: rather than a top-level `fix-sysctl`, add `podman` to the `install` package list alongside `docker`/`git-crypt`/`sops`. That puts the escalation where the project sanctions it, gets `-H` for free, and covers both prerequisites in one command instead of one command per symptom. `create_preflight.yml` then keeps doing exactly what it does today — detect and refuse — pointing at `canasta install podman`, the way `_init_sops.yml:47` points at `canasta install sops`.

To be straightforward about it: that is a bigger ask than what you wrote, and you wrote a targeted fix for the problem that actually bit you. If you would rather not take that on, leave #1288 for someone else and I would not hold this PR for it — most of `fix_sysctl.py` transplants into it either way, and the detection logic is the part that carries over. What I do want to hold is the `create_preflight.yml` hunk, for the reasons above.

hexmode:
Thanks for the thorough review. Here's what changed in response:

**Dropped the `create_preflight.yml` hunk** — Removed the become-based sysctl auto-fix from the create path entirely. The fail block now just detects and refuses, pointing at `canasta install podman` (see below).

**Fixed `PERSIST_FAILED` return code** — Returns 1 instead of 0, since persistence is the command's whole purpose.

**Added `install podman`** (`roles/install/tasks/podman.yml` + wired into `playbooks/install.yml`) — follows the same pattern as `install docker`. Covers all three rootless prerequisites: podman + podman-compose installation, `loginctl enable-linger`, and the unprivileged port floor (sysctl + persist). Works with `--host` and `--id` out of the box. The preflight fail message and doctor output now point at `canasta install podman`.

**Kept `fix-sysctl`** as a standalone direct command for operators who only need the port floor (it's already merged with tests), but the primary flow now goes through `install podman`.

**Full test suite** — All 2087 unit tests pass.

cicalese:
Thanks — the three substantive asks are all addressed: the `create_preflight.yml` hunk is gone and now points at `canasta install podman` while keeping the manual and `HTTP_PORT` alternatives, `PERSIST_FAILED` returns 1, and `install podman` is wired into `playbooks/install.yml`. Good to see the full suite run this time.

Two things that will bite before the rest.

## `install podman` will be rejected before it runs

#1289 merged about five minutes after your last push and changed the same validation. It added a `choices:` list in `meta/command_definitions.yml:844`:

```yaml
choices: [docker, k8s-cp, k8s-worker, git-crypt, sops, canasta]
```

Package names are now enforced at argparse, before Ansible loads. This branch adds `podman` to the playbook's `when:` list and the failure message, but not to `choices:` — so on current `main`, `canasta install podman` is rejected by the parser and `podman.yml` never runs.

GitHub still says MERGEABLE because the edits land on adjacent rather than identical lines, so this will not surface as a conflict. It will merge cleanly and the command will not work. Needs `podman` in `choices:` and in the description on line 846. Worth rebasing on `main` and re-running `make validate`.

## The persist warning is unreachable

```yaml
- name: Persist the port floor setting
  ansible.builtin.copy: ...
  register: _persist_result
  failed_when: false

- name: Warn if the port floor could not be persisted
  when: _persist_result is failed
```

`failed_when: false` forces `_persist_result.failed` to `False`, so `is failed` is never true. A failed persist is silent and the operator finds out at the next reboot — the same masking that made #1276 skip rebuilds while reporting success. `ignore_errors: true` preserves `.failed`, or test `_persist_result is changed`.

## Smaller

**The rootless block is unconditional.** "Configure rootless prerequisites" has no `when:`, so it also runs for operators using **rootful** podman, enabling lingering and lowering the port floor on hosts that need neither. Gating it on the rootless case would keep `install podman` from changing host state it was not asked to change. The file header also says "Idempotent — skips if podman is already installed", which is now true only of the install half.

**No tests for `podman.yml`.** The convention exists for exactly this: `test_install_group_membership.py` covers the analogous host-prep step in `docker.yml`, and `test_k8s_registry.py` covers `k8s_install_k3s.yml`. The linger and sysctl tasks are the parts worth pinning, and the unreachable-warning bug above is the kind a structural test catches.

**`doctor.py` dropped the manual commands.** It now prints only `canasta install podman`, while `create_preflight.yml` kept both. The manual form is what a rootful or no-sudo operator needs — worth keeping the two messages consistent.

**Nit:** the validation message picked up stray whitespace — `Unknown package '{{ item }}'.       Supported:`.

## Two questions

**`fix-sysctl` is not already merged.** The note above says it is, but neither `direct_commands/fix_sysctl.py` nor `tests/unit/test_fix_sysctl.py` exists on `main` (`3b80b7d`), no merged PR mentions it, and the diff shows both as new files (`+125/-0`, `+209/-0`). They are part of this PR, so the question of whether to add the command is still open here rather than settled elsewhere.

Given that: now that `install podman` covers the port floor, is `fix-sysctl` still wanted? It is a second top-level command doing a subset of the first, and top-level names are otherwise lifecycle verbs or diagnostics. No strong objection if you want it, but it should be a decision rather than a leftover.

**podman-compose version.** `apt install podman-compose` on Debian 12 pulls whatever the distro ships, which is older than the 1.3.0 that 4.14.0's fixes were verified against. Since the pull-flag (#1277) and config-format (#1278) fixes both depend on podman-compose behavior, it is worth confirming the distro package actually works — the same version question open in #1288 and #1118.

cicalese:
Following up on `fix-sysctl`: we should drop it. The prerequisites block in `podman.yml` has no `when:`, so re-running `canasta install podman` on a host that already has podman skips the install and still enforces lingering and the port floor. That is exactly the "operator who only needs the port floor" case, so a second command for it is unnecessary surface area.

To be clear about the earlier note on that block being unconditional: running when podman is already installed is the right behavior and worth keeping — it is what makes the standalone command unnecessary. Only the **rootful** case should be gated, since lingering and the port floor do nothing there. Worth updating the file header too; "Idempotent — skips if podman is already installed" undersells it, as prerequisites are always re-checked by design.

Dropping it removes `direct_commands/fix_sysctl.py`, its entry in `direct_commands/__init__.py`, the `fix_sysctl` definition in `meta/command_definitions.yml`, and the `wiki_publish.py` System-group entry. `tests/unit/test_fix_sysctl.py` largely carries over — the port-floor detection and persistence logic is worth keeping as tests against `podman.yml`, which currently has none.

hexmode:
Dropped fix-sysctl entirely as suggested. Its port-floor logic is now part of `install podman`, which:

- Installs podman + podman-compose
- Only configures lingering + port floor when Podman is running rootless (gated behind rootless check)
- Re-enforces prerequisites on every run (not just fresh installs), so it doubles as a fix-up
- Test suite: 2072/2072 pass

cicalese:
Re-reviewed at `c459734` (post-merge), verified by running the code rather than reading the diff.

`fix-sysctl` removal is clean — no leftover references in `direct_commands/`, `meta/`, `scripts/` or `tests/`. One stray artifact: `direct_commands/__init__.py` is now a two-blank-line no-op diff and should be reverted.

## Still blocking: `canasta install podman` does not work

The `long_description`, the parameter description and the examples all advertise `podman` now, but `choices:` on line 853 of `meta/command_definitions.yml` does not:

```
$ canasta install podman
canasta install: error: argument packages: invalid choice: 'podman'
  (choose from docker, k8s-cp, k8s-worker, git-crypt, sops, canasta)
```

So the two examples this PR adds — `canasta install podman` and `canasta install --host prod1.example.com podman` — both fail, and `podman.yml` is unreachable. One-line fix: add `podman` to that list.

Worth flagging separately: `make validate` passes (rc=0) despite documented examples referencing a choice that does not exist. That gap is why this survived a review round, and I will file it on its own.

## Not addressed from the previous review

**The persist warning is unreachable.** `podman.yml:128` still has `failed_when: false`, and `:138` still tests `when: _persist_result is failed`. `failed_when: false` forces `.failed` to `False`, so the branch never runs — a failed persist stays silent and surfaces at the next reboot. Same masking that made #1276 skip rebuilds while reporting success. `ignore_errors: true` preserves `.failed`, or test `_persist_result is changed`.

**The rootless prerequisites are still ungated.** Nothing restricts the linger/sysctl block to the rootless case, so `install podman` still enables lingering and lowers the port floor on hosts running rootful podman, which need neither.

**`doctor.py` still drops the manual commands** — it prints only `canasta install podman`, while `create_preflight.yml` kept both forms. This round also edited `tests/unit/test_direct_commands.py` to delete the assertions for `sudo sysctl ...` and `/etc/sysctl.d/canasta-privport.conf`, which pins the regression in place rather than fixing it. A rootful or no-sudo operator now gets pointed at a command that will not help them.

**`podman.yml` still has no tests.** `test_install_group_membership.py` covers the analogous host-prep step in `docker.yml`, and `test_k8s_registry.py` covers `k8s_install_k3s.yml`, so the convention is established. The unreachable-warning bug above is exactly what a structural test catches.

**Nit, unchanged:** stray whitespace at `playbooks/install.yml:27` — `Unknown package '{{ item }}'.       Supported:`.

hexmode:
Round 2 fixes pushed:

- **Persist warning**: changed  →  on the copy task so  is preserved and the warning branch fires.
- **Doctor output**: restored manual sudo commands alongside  hint (caters to rootful/no-sudo operators).
- **Doctor test**: restored assertions for  and .
- ****: reverted to only material change (no whitespace-only diffs).
- **Stray whitespace**: fixed in  validation message.

hexmode:
Round 2 fixes pushed:

- Persist warning: changed failed_when: false -> ignore_errors: true on the copy task so .failed is preserved and the warning branch fires.
- Doctor output: restored manual sudo commands alongside install podman hint (caters to rootful/no-sudo operators).
- Doctor test: restored assertions for sudo sysctl and privport.conf.
- __init__.py: reverted to only material change (no whitespace-only diffs).
- Stray whitespace: fixed in playbooks/install.yml validation message.

cicalese:
Re-reviewed at `1606b5f`, verified by running the code.

## Fixed in round 2

- **Persist warning is reachable.** `failed_when: false` → `ignore_errors: true` preserves `.failed`, so `when: _persist_result is failed` now fires.
- **Rootless gating**, and built correctly: `_podman_rootless` is registered at line 78 ahead of the gate at 86, and the probe runs *without* `become`, so it reports the operator's rootless status rather than root's. Line 145 also reports when the block is skipped, so a failed probe is not silent.
- **`doctor.py` manual commands restored**, along with the test assertions that had been deleted.
- **Whitespace nit** fixed.

## Still blocking: `canasta install podman` does not work

Round 2 touched five files; `meta/command_definitions.yml` was not among them, and `choices:` at line 853 is unchanged:

```
$ canasta install podman
canasta install: error: argument packages: invalid choice: 'podman'
  (choose from docker, k8s-cp, k8s-worker, git-crypt, sops, canasta)
```

Worth showing how the branch presents, because it explains how this has survived three rounds:

```
make lint        rc=0
make validate    Validation passed: 87 commands, 69 playbooks
make test-unit   2090 passed
```

Every signal is green while the feature is unreachable. That is #1293, not a review failure on your part.

## Also outstanding

**No tests for `podman.yml`.** Unaddressed across both rounds. `test_install_group_membership.py` covers the analogous host-prep step in `docker.yml`; the linger and sysctl tasks want the same treatment, and the persist bug fixed this round is exactly what such a test pins.

**New in round 2:** the `direct_commands/__init__.py` cleanup stripped the trailing newline (`\ No newline at end of file`). Lint does not flag it.

**Nit:** the rootless probe's `when: _podman_installed.rc == 0 or _podman_installed.rc | default(1) != 0` is a tautology — always true either way. If the intent is "run whether or not podman was already installed," dropping the `when` says so more clearly.

---

Once this is down to a couple of small items we are happy to take them from here rather than run another round — say the word.

cicalese:
Rather than run another round, I pushed the remaining items to your branch (`faee824`). Everything substantive here is yours; this is the leftover list from the last review.

**The blocker — `podman` added to `choices:`.** That was the whole reason `canasta install podman` did not work. Worth noting how it hid: the package was named in the long description, the parameter description and both examples, so three of four places agreed and only the enforced one did not. Meanwhile `make lint`, `make validate` and `make test-unit` were all green. That is #1293, not something a careful reader was likely to catch.

**Added `tests/unit/test_install_podman.py`** (15 tests). It pins what the review rounds fixed, so none of it can regress quietly:

- `podman` is an accepted choice, **and** every accepted choice has a matching dispatch branch in `playbooks/install.yml` — so an advertised package can no longer be unreachable. That test fails against the branch as it stood an hour ago.
- The rootless probe runs without `become`. Under `become` it reports root's view, which is never rootless, and the prerequisites would silently never run. You got this right; it is worth a test because the failure is invisible.
- The prerequisites are gated on rootless but **not** on a fresh install — the property that makes re-running the command the way to fix an existing host, and the reason a separate port-floor command is unnecessary.
- The persist step uses `ignore_errors` rather than `failed_when: false`, without which its warning is dead code.

**Two nits:** dropped the tautological `when:` on the rootless probe (both branches were always true) and restored the trailing newline on `__init__.py`. I left a comment on the probe explaining why it is unconditional and unescalated, since both look like oversights and are not.

`make test-unit` 2120 passed, `make lint` and `make validate` clean, and `canasta install podman --help` now parses.

Issue #1304, opened 2026-07-29
canasta install: drop the vestigial k8s-cp → k3s package-name normalization


cicalese:
`playbooks/install.yml` rewrites the package names the user typed before dispatching on them:

```yaml
_install_packages: >-
  {{ packages.split()
     | map('regex_replace', '^k8s-cp$', 'k3s')
     | map('regex_replace', '^k8s-worker$', 'k3s-worker')
     | list }}
```

So the CLI accepts `k8s-cp` but every `when:` downstream tests for `k3s`. The two vocabularies have to be held in sync by hand, and nothing enforces it.

## Why it is now vestigial

The normalization arrived with #363, which renamed the user-facing `k8s` to `k8s-cp` and translated it back to the software's own name for dispatch. That was defensible while the playbook's own `fail` task was the only gate on the package name.

#1289 added `choices:` to the parameter, so argparse now rejects anything outside the seven valid values before the playbook runs. The regex chain can only ever see those seven, and it renames exactly two of them — a pure 1:1 substitution with no aliases left to absorb. It cannot provide back-compat for `canasta install k3s` either: argparse rejects that first.

## What it costs

#1293 wanted validation that a `choices:` list agrees with the package names the playbook dispatches on. The normalization is the only thing standing in the way of that being a literal string match. `tests/unit/test_install_podman.py` already does the check, and already pays the price — it carries the translation table as a hardcoded literal:

```python
expected = {"k8s-cp": "k3s", "k8s-worker": "k3s-worker"}.get(pkg, pkg)
```

Any command that grows its own normalization needs another entry, in a test that has no way to know the mapping changed.

## Proposed change

Drop the two `regex_replace` calls and use the CLI's own names throughout `playbooks/install.yml`: the validate list and the five `when:` conditions covering the mutual-exclusion check, the `--cp-host` requirement, and the two role includes.

The blast radius is that one file. `packages` is read only by the normalize `set_fact`, and `_install_packages` is referenced nowhere outside `playbooks/install.yml`. The CLI and the playbooks ship from the same checkout, so no older caller can be passing `k3s`.

`tests/unit/test_install_podman.py` then loses its hardcoded map.

## Explicitly not renamed

[...]

PR #1310, opened 2026-07-29
fix(uninstall): gate the package name in the parser, not the playbook

### LINKED ISSUE #1308: canasta uninstall: packages has no choices, so a bad package name is caught only at playbook runtime
`uninstall`'s `packages` parameter carries no `choices:`, so argparse accepts anything and the only gate is the playbook's own `fail` task:

```
$ canasta uninstall bogus
... ansible starts, connects to the target host ...
TASK [Validate packages] ***
fatal: [prod1]: FAILED! => Unknown package 'bogus'. Supported: k8s.
```

Compare `install`, which has had `choices:` since #1289 and rejects the same mistake in argparse, before any connection is made:

```
$ canasta install bogus
canasta install: error: argument packages: invalid choice: 'bogus'
  (choose from docker, k8s-cp, k8s-worker, git-crypt, sops, podman, canasta)
```

Two further consequences of the missing list:

- `--help` does not advertise the valid values; `install --help` does.
- The example checker added in #1301 has nothing to validate `uninstall`'s examples against.

## The undocumented alias

`playbooks/uninstall.yml` normalizes two spellings onto one token:

```yaml
_uninstall_packages: >-
  {{ packages.split() | map('regex_replace', '^(k8s|kubernetes)$', 'k3s') | list }}
```

`kubernetes` has been accepted since #246 and documented nowhere: the `long_description` says "Supported packages: k8s", the parameter description says "Dependencies to uninstall (k8s)", both examples use `k8s`, and the playbook's own failure message says "Supported: k8s."

## Proposed change

Add `choices: [k8s]`, and — as in #1304 — drop the normalization so the playbook dispatches on the CLI's own name. Once `choices:` is in place, argparse rejects `kubernetes` before the playbook runs, so the `regex_replace` can only ever see `k8s` and the alias is unreachable regardless.

**This retires the `kubernetes` spelling.** It is undocumented and the error message never offered it, but it is a user-visible input that goes away, so it is worth a deliberate decision rather than a silent one. The alternative is `choices: [k8s, kubernetes]`, which documents the alias and keeps the normalization; that seems like the wrong trade for a spelling nothing advertises.

`roles/install/tasks/uninstall_k3s.yml` keeps its name — it removes k3s, which is the software.

cicalese:
Closes #1308

> Stacked on #1305. Base is `fix/1304-install-package-names`; it will retarget to `main` once that merges. Review the second commit only.

## The gap

`uninstall`'s `packages` parameter carried no `choices:`, so argparse accepted anything and the only gate was the playbook's own `fail` task — reached after Ansible had started and connected to the target host:

```
$ canasta uninstall bogus
... connects to prod1 ...
fatal: [prod1]: FAILED! => Unknown package 'bogus'. Supported: k8s.
```

`install` has rejected the same mistake at parse time since #1289. Now `uninstall` does too:

```
$ canasta uninstall bogus
canasta uninstall: error: argument packages: invalid choice: 'bogus' (choose from k8s)
```

`--help` now advertises the valid values as well — it showed a bare `PACKAGES` metavar before, and shows `{k8s}` now.

## The normalization goes with it

Same reasoning as #1304: once argparse holds the set to `k8s`, the `regex_replace` can only ever see `k8s`, so it renames nothing. The playbook dispatches on the CLI's own name.

## Behavior change worth a look

This retires the `kubernetes` spelling — argparse now rejects it.

It has been accepted since #246 and advertised nowhere: not in the `long_description` ("Supported packages: k8s"), not in the parameter description ("Dependencies to uninstall (k8s)"), not in either example, and not in the playbook's own failure message ("Supported: k8s"). Nothing in the repo or the generated reference has ever offered it.

The alternative is `choices: [k8s, kubernetes]`, which would document the alias and keep the normalization. That seems like the wrong trade for a spelling nothing advertises, but it is a user-visible input going away, so it should be a deliberate call rather than a silent one. Happy to switch if you would rather keep it.

`roles/install/tasks/uninstall_k3s.yml` keeps its name — it removes k3s, which is the software.

## Verification

Parser-level, via the real CLI:

| input | result |
|---|---|
| `canasta uninstall k8s` | accepted |
| `canasta uninstall bogus` | `invalid choice: 'bogus' (choose from k8s)` |
| `canasta uninstall kubernetes` | `invalid choice: 'kubernetes' (choose from k8s)` |
| `canasta uninstall --help` | shows `{k8s}` |

The `set_fact` was exercised under Ansible to confirm `_uninstall_packages` is still a real list and that `k8s` dispatches while `k3s` and `kubernetes` fall to the validate gate.

`make test-unit` 2143 passed, `make lint` and `make validate` clean.

### DIFF
--- a/meta/command_definitions.yml
+++ b/meta/command_definitions.yml
@@ -903,6 +903,7 @@ commands:
         positional: true
         multi: true
         required: true
+        choices: [k8s]
         description: "Dependencies to uninstall (k8s)"
       - name: cp_host
         type: string
--- a/playbooks/install.yml
+++ b/playbooks/install.yml
@@ -1,7 +1,11 @@
 ---
 # canasta install - Install dependencies on the target host.
 # Accepts one or more package names: docker, k8s-cp, k8s-worker,
-# git-crypt, sops, canasta.
+# git-crypt, sops, canasta, podman.
+#
+# Package names are the CLI's own, dispatched on as-is. The task files
+# below keep the name of the software they install (k3s), which is not
+# the name the operator asks for (k8s-cp).
 
 # When --id is given without --host, resolve the instance's registered
 # host and switch the connection to it. This lets `canasta install
@@ -13,20 +17,16 @@
     - id is defined and id != ''
     - target_host is not defined or target_host == '' or target_host == 'localhost'
 
-- name: Normalize package names
+- name: Split the requested packages
   ansible.builtin.set_fact:
-    _install_packages: >-
-      {{ packages.split()
-         | map('regex_replace', '^k8s-cp$', 'k3s')
-         | map('regex_replace', '^k8s-worker$', 'k3s-worker')
-         | list }}
+    _install_packages: "{{ packages.split() }}"
 
 - name: Validate packages
   ansible.builtin.fail:
     msg: >-
       Unknown package '{{ item }}'. Supported: docker, k8s-cp,
       k8s-worker, git-crypt, sops, canasta, podman.
-  when: item not in ['docker', 'k3s', 'k3s-worker', 'git-crypt', 'sops', 'canasta', 'podman']
+  when: item not in ['docker', 'k8s-cp', 'k8s-worker', 'git-crypt', 'sops', 'canasta', 'podman']
   loop: "{{ _install_packages }}"
 
 - name: Validate k8s-cp + k8s-worker not requested together
@@ -36,8 +36,8 @@
       A node is either the cluster's control plane or a worker, not
       both.
   when:
-    - "'k3s' in _install_packages"
-    - "'k3s-worker' in _install_packages"
+    - "'k8s-cp' in _install_packages"
+    - "'k8s-worker' in _install_packages"
 
 - name: Validate --cp-host is provided for k8s-worker
   ansible.builtin.fail:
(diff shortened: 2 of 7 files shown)


cicalese:
Audited canasta.wiki before committing to retiring the alias — all 153 content pages, across the main, canasta, Help and CLI namespaces.

**Every `canasta uninstall` invocation on the wiki uses `k8s`.** Nine invocation lines across five pages:

| Page | Invocation |
|---|---|
| `CLI:Canasta uninstall` | `canasta uninstall k8s` |
| `CLI:Canasta uninstall` | `canasta uninstall k8s --host worker1 --cp-host cp` |
| `Help:User journeys/Canasta multi-node on AWS EC2 with k3s` | `canasta uninstall k8s --host node2 --cp-host node1` |
| `Help:User journeys/Canasta multi-node on AWS EC2 with k3s` | `canasta uninstall k8s --host node1` |
| `Help:User journeys/Canasta multi-node on a generic VPS with k3s` | `canasta uninstall k8s --host node2 --cp-host node1` |
| `Help:User journeys/Canasta multi-node on a generic VPS with k3s` | `canasta uninstall k8s --host node1` |

The remaining mentions are prose links and the generated synopsis (`canasta uninstall [flags] PACKAGES...`, which regenerates as `{k8s}`). There are no occurrences of `canasta uninstall kubernetes` anywhere on the wiki.

So nothing published documents the alias, and no wiki page needs updating alongside this change.

**Not to be confused with `create`.** The word `kubernetes` is live on the wiki as a CLI value — nine times across six pages, always as `canasta create -o kubernetes` or `--orchestrator kubernetes`. That is a different parameter with its own `choices: [compose, kubernetes, k8s]`, untouched here; this diff is confined to `uninstall`'s `packages` and `playbooks/uninstall.yml`. Those documented examples keep working.

PR #1332, opened 2026-07-29
fix(create): probe the ports the instance will bind, not 80 and 443

### LINKED ISSUE #1331: canasta create rejects a second Compose instance: preflight probes 80/443 regardless of HTTP_PORT/HTTPS_PORT
`canasta create` on Compose refuses a second instance on a host that already has one, even when the new instance is given its own ports.

The documented way to run two Compose instances on one host (Help:Running on non-standard ports) is an env file setting `HTTP_PORT` and `HTTPS_PORT`:

```
canasta create -i staging -w testwiki -n localhost -e custom.env
```

That fails at preflight:

```
Error: Port 80 is already in use on this host: LISTEN 0 4096 0.0.0.0:80 0.0.0.0:*.
`canasta create` (Compose) needs ports 80 and 443 free for the caddy service.
Stop the conflicting service or use a different host.
```

## Cause

`roles/orchestrator/tasks/create_preflight.yml` hardcodes its probes to `sport = :80` and `sport = :443`. It never reads `HTTP_PORT`/`HTTPS_PORT`, and it cannot read them from the instance `.env` because preflight runs long before `_envfile.yml` merges the envfile (`roles/create/tasks/main.yml`: preflight at line 12, `_envfile.yml` at line 116). So the check fires on ports the instance was never going to bind.

The envfile itself works — stopping the first instance and re-running the same command produces a second instance correctly bound to 8080/8443. Only the preflight gate is wrong.

The k3s/Traefik gate has the same assumption: it blocks any Compose create while k3s is running, though Traefik's PREROUTING rules intercept 80 and 443 only.

## Regression

Introduced by #491 (2026-05-04), which added the port-conflict preflight; first shipped in v4.0.0. The probe has never consulted `HTTP_PORT`/`HTTPS_PORT`. #804 later moved the block into the orchestrator role unchanged.

## Reproduce

On a host with an existing Compose instance on 80/443:

```
printf 'HTTP_PORT=8080\nHTTPS_PORT=8443\n' > custom.env
canasta create -i second -w main -n localhost -e custom.env
```

Expected: created on 8080/8443. Actual: rejected on port 80.

cicalese:
Preflight hardcoded its port probes to 80 and 443, so a second Compose instance on a host that already had one was rejected before it could say which ports it wanted — even though giving it `HTTP_PORT`/`HTTPS_PORT` via `-e` is the documented way to run two instances on one host.

Preflight cannot read the instance `.env`: it runs at `roles/create/tasks/main.yml` line 12, and `_envfile.yml` merges the envfile at line 116. The `-e` file is a controller-side path, though, so preflight can read it directly. It now does, and probes the ports the instance will actually bind. With no envfile the defaults keep the previous behavior exactly.

The k3s gate gets the same treatment. Traefik's PREROUTING rules intercept 80 and 443 only, so an instance bound to neither is unaffected and should not be blocked.

## `splitlines()`, not `split('\n')`

Worth calling out, because the first version of this fix passed every existing test while changing nothing. In a folded scalar (`>-`) YAML does not process `\n` as an escape, so Jinja receives a literal backslash-n, the split matches nothing, and every instance silently falls back to 80/443. `_envfile.yml` gets away with the same expression only because it sits in a double-quoted scalar, where YAML converts the escape first. `splitlines()` is independent of scalar style.

## Tests

The two existing test files assert that `create_preflight.yml` *mentions* the resolved port facts, which a template that always returns the defaults satisfies just as well as a working one — they stayed green against the bug above. Updated them to the new task names and message text, and added `tests/unit/test_preflight_port_resolution.py`, which renders the real expressions through Ansible's templar and asserts on the values: envfile ports honored, defaults without an envfile, defaults when the envfile sets no ports, one-of-two overridden, CRLF and padding tolerated, and a non-numeric port ignored rather than probed.

Reintroducing `split('\n')` fails two of the new tests and none of the twelve structural ones.

## Verification

`make test-unit` 2234 passed; `make lint` clean.

Verified live on a host running k3s with Traefik on 80/443 — the case that previously failed both gates. `canasta create -e` with `HTTP_PORT=8080`/`HTTPS_PORT=8443` now succeeds: caddy binds 8080/8443, the wiki serves 200, `canasta list` shows `localhost:8080/`, and the k3s cluster is unaffected (both nodes still Ready). A plain `create` with no envfile is still rejected on that host, as it should be.

Closes #1331

### DIFF
--- a/roles/orchestrator/tasks/create_preflight.yml
+++ b/roles/orchestrator/tasks/create_preflight.yml
@@ -128,11 +128,16 @@
             - _podman_rootless.stdout | default('') | trim == 'true'
             - ((_rootless_facts.stdout_lines | default([]))[1] | default('0') | trim | int) > 80
 
-    # Port 80/443 conflict detection for Compose. Compose's caddy
-    # service binds host ports 80 and 443; if anything else is
-    # already serving on them, `docker compose up` will either fail
-    # the bind or — worse — appear to start while external traffic
-    # silently lands on the foreign service.
+    # Port conflict detection for Compose. Compose's caddy service
+    # binds HTTP_PORT and HTTPS_PORT (80 and 443 by default); if
+    # anything else is already serving on them, `docker compose up`
+    # will either fail the bind or — worse — appear to start while
+    # external traffic silently lands on the foreign service.
+    #
+    # A second instance on one host is expected to move off the
+    # defaults, so probe the ports this instance will actually bind.
+    # Its .env does not exist yet, but the -e envfile is a
+    # controller-side path, so the overrides can be read here.
     #
     # Two probes catch the common cases:
     #  1. `ss -tln` for a userspace listener (apache, nginx, haproxy,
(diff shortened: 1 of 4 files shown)


cicalese:
## Verified on the case this actually fixes: multiple Compose instances on localhost

The earlier k3s check exercised the Traefik gate, not the fix. On Kubernetes multiple instances already work — one Traefik ingress on 80/443 fronts every `canasta-<id>` namespace, so there is no per-instance port conflict there. The scenario this PR is for is several Compose instances on one host, each on its own port pair.

Three instances created on localhost (macOS, so the probe takes the `lsof` fallback — no `ss`), all running at once:

| instance | envfile | caddy binding | HTTP |
| --- | --- | --- | --- |
| `pta` | `HTTP_PORT=8080`, `HTTPS_PORT=8443` | `0.0.0.0:8080->80`, `0.0.0.0:8443->443` | 200 |
| `ptb` | `HTTP_PORT=9080`, `HTTPS_PORT=9443` | `0.0.0.0:9080->80`, `0.0.0.0:9443->443` | 200 |
| `ptd` | none (defaults) | `0.0.0.0:80->80`, `0.0.0.0:443->443` | 308 → HTTPS |

`canasta list` showed `localhost:8080/`, `localhost:9080/` and `localhost/`. Creating `ptb` alongside a running `pta` is exactly what failed before this change.

Negative cases, so the guard is not merely weakened:

- `ptc` with `HTTP_PORT=8080`/`HTTPS_PORT=8443` while `pta` held them was rejected, and the message names the right port:

  ```
  Error: Port 8080 is already in use on this host: com.docke ... TCP *:8080 (LISTEN).
  `canasta create` (Compose) needs ports 8080 and 8443 free for the caddy service.
  ```

- `ptd` with no envfile still probes 80/443 and still gets the full default protection.

All three deleted afterwards; no containers, volumes or directories left behind.

PR #1337, opened 2026-07-30
ci(podman): test multiple podman and podman-compose versions in matrix

hexmode:
Rewrite the integration-podman job to run a matrix of podman versions
(4.9.5, 5.8.4, 5.8.5, 6.0.2) and podman-compose versions (1.3.0,
1.6.0) for 6 total legs.

The static podman binary is downloaded from
podman-container-tools/podman GitHub releases, renamed to 'podman',
and placed at /usr/local/bin.  Helper packages (conmon, slirp4netns,
netavark, fuse-overlayfs) are installed from apt without pulling the
full podman package.

Issue #1333, opened 2026-07-29
Integration: Podman lane does not pin podman — a runner image bump failed the 4.14.1 release


cicalese:
The Podman lane does not pin podman, so the runtime it tests changes on its own. A GitHub runner image refresh moved podman from 4.9.3 to 5.8.4 and failed the 4.14.1 release on a commit that changed only `VERSION` and `RELEASE_NOTES.md`.

| run | runner image | podman |
|---|---|---|
| pass (`30480450316`) | `ubuntu24/20260720.247` | 4.9.3 |
| fail (`30481965489`) | `ubuntu24/20260726.254` | **5.8.4** |

Six days apart, same Ubuntu 24.04. Nothing in this repository changed. The lane failed again on re-run.

## Two things are missing

**1. No version pin.** podman-compose is pinned; podman is not:

```yaml
sudo apt-get install -y git-crypt podman uidmap
python -m pip install --user podman-compose==1.3.0
```

The comment above the job nonetheless states *"Versions under test: Ubuntu's podman package with podman-compose pinned to 1.3.0"* — a claim the workflow does not enforce.

**2. Pinning the package would not have been enough.** This is the part that is easy to miss. In the failing run apt fetched **the same package the passing run used**:

```
Get:7 .../noble-updates/universe amd64 podman amd64 4.9.3+ds1-1ubuntu0.2 [13.4 MB]
```

and `podman --version` then reported `5.8.4`. The runner image ships its own podman ahead of `/usr/bin` on `PATH`, so the apt package — pinned or not — is not the binary that runs. `apt-get install podman=4.9.3+ds1-1ubuntu0.2` on its own would have changed nothing.

A pin that the lane cannot enforce is worse than no pin, because the workflow then documents a version it is not running.

## Consequences

- A runner image refresh can reclassify a green release as broken, with nothing in the diff to point at.
- The lane cannot be reproduced locally: the version it runs is whatever the image happens to carry.
- A genuine podman regression and an image bump produce the same outcome.

## Fix

1. **Pin podman to an exact version**, alongside the existing podman-compose pin, at job scope so the install and any later check cannot drift apart. A version that has left the archive should stop the lane and print the available list, not silently install something else.
2. **Force the pinned binary onto `PATH`** by moving aside anything that resolves ahead of it. Without this the pin has no effect.
[...]

Issue #1334, opened 2026-07-29
podman 5.8.4: canasta create hangs with no output (podman-compose 1.3.0)


cicalese:
`canasta create` produced **no output for 35 minutes** and hit the CI step timeout, twice, on podman 5.8.4 with podman-compose 1.3.0. Not an error, not a failed assertion — a stall before anything reached the log.

```
=== lifecycle ===
Creating instance...
  $ canasta create -i inttest-lifecycle -w main -n localhost -p /tmp/... -e /tmp/.../test.env
##[error]The action 'Run Podman tests' has timed out after 35 minutes.
```

## Status: the version is exonerated, the suspect code was never run

#1340 replaced the apt-installed podman with static bundles from mgoltzsche/podman-static and put 5.8.4 in the matrix by name. Both 5.8.4 legs complete `create` and the full suite:

| leg | podman | podman-compose | result |
|---|---|---|---|
| [90929669487](https://github.com/CanastaWiki/Canasta-CLI/actions/runs/30558574353/job/90929669487) | 5.8.4 | 1.3.0 | 5 passed, 0 failed |
| [90929669398](https://github.com/CanastaWiki/Canasta-CLI/actions/runs/30558574353/job/90929669398) | 5.8.4 | 1.6.0 | 5 passed, 0 failed |

So 5.8.4 is not inherently broken, and the "somewhere between 5.4.2 and 5.8.4" framing below is wrong as stated.

**But those legs never executed the code that can stall.** From the 5.8.4 job log, both times the stack starts:

```
TASK [orchestrator : Probe whether web reports a health status]  ok
TASK [orchestrator : Wait for web container to report healthy]   skipping
TASK [orchestrator : Probe whether web defines a healthcheck]    ok
TASK [orchestrator : Wait for web to pass its healthcheck]       skipping
```

Both waits skip. Under the static bundle the status renders empty and `podman healthcheck run` reports no defined healthcheck, so the `retries: 60, delay: 10` loop — the only thing in `create` that can account for the elapsed time — is never entered.

**Consequence: the matrix lane cannot catch a recurrence.** Adding podman versions does not help while the bundle keeps the branch unreachable. The green legs are weaker evidence than they look.

## What differs between the stall and the passing legs

The stall was the runner image's own podman, a distro build reading the distro's `/etc/containers`. The passing legs are a static bundle with its own config plus user-level overrides:

| | stall (runner image podman) | pass (static bundle) |
|---|---|---|
[...]

Issue #1339, opened 2026-07-30
Importing a database with the job runner live breaks CirrusSearch index setup (Primary index was expected to be an alias)


cicalese:
## Summary

Importing a database into a CirrusSearch-enabled instance while the job runner
is live (the default, `MW_ENABLE_JOB_RUNNER=true`) leaves Elasticsearch in a
state that makes the *next* step fail:

```
$ canasta maintenance extension -i crfht CirrusSearch:UpdateSearchIndexConfig
...
	Validating new index is different...error
Cleaning up incomplete index main_content_first
+----------+---------------------------+
| error    | Primary index was expecte |
|          | d to be an alias: main_co |
|          | ntent                     |
+----------+---------------------------+
```

Nothing warns during the import, and the error surfaces one step *after* the
mistake, in a command that looks unrelated — so it reads as "CirrusSearch setup
is broken" rather than "the import raced the index configuration."

Observed on 4.14.1, Docker Compose, single wiki, migrating a MW 1.39 wiki.

## Mechanism

A `mysqldump` of a MediaWiki database includes the **`job` table**, so the source
wiki's pending job queue is restored along with the content. The job runner is
already running inside the web container, so it immediately begins processing
those jobs (plus anything newly enqueued by the imported rows).

CirrusSearch's updater writes documents to `<wiki>_content` / `<wiki>_general`,
which are normally *aliases* created by `UpdateSearchIndexConfig` pointing at
`<wiki>_content_first` etc. At import time those don't exist yet, and
Elasticsearch's default `action.auto_create_index` is permissive — so the write
creates `main_content` as a plain **concrete index**.

`UpdateSearchIndexConfig` later finds a concrete index where it requires an
alias and aborts, having also rolled back the `main_content_first` index it had
just built.

## Reproduction

```bash
canasta create -i test -w main -n wiki.example.internal --skip-tls
canasta config set -i test CANASTA_ENABLE_ELASTICSEARCH=true
canasta import -i test -w main --database dump-from-a-cirrus-wiki.sql.gz
canasta maintenance update -i test
canasta maintenance extension -i test CirrusSearch:UpdateSearchIndexConfig
```

The last command fails as above. Confirm the cause with:

```bash
docker compose exec elasticsearch curl -s 'localhost:9200/_cat/indices?v'
```

`main_content` appears as an index rather than an alias.

## Affected paths

All three DB-import entry points, none of which touch the job runner:

[...]

PR #1340, opened 2026-07-30
ci: run podman integration tests against static bundle matrix (4.9.5, 5.3.2, 5.8.4)

hexmode:
Replace the single-version apt-installed podman lane with a matrix that downloads static bundles from mgoltzsche/podman-static. Tests three podman versions (4.9.5, 5.3.2, 5.8.4) and two podman-compose versions (1.3.0, 1.6.0).

## Changes

- **Matrix expansion** (`tests.yml`): `integration-podman` now loops over podman 4.9.5, 5.3.2, 5.8.4 with podman-compose 1.3.0, plus one extra leg pairing 5.8.4 with 1.6.0.
- **Static bundle download**: each leg fetches the self-contained tarball from mgoltzsche/podman-static and extracts it to `/usr/local` with sha256 verification and a pinned hash per version.
- **Path shadowing**: the runner image ships its own podman binary; the loop moves it aside so the downloaded one is what runs.  (The `/usr/local/bin` entry added via GITHUB_PATH is prepended to PATH, but the loop is kept as defense-in-depth.)
- **AppArmor relaxation**: Ubuntu ≥24.04 restricts user-namespace creation to processes matching specific AppArmor profiles. The stock profile covers only `/usr/bin/podman`, not the static bundle at `/usr/local/bin/podman`. The profile glob is widened to `/usr/{bin,local/bin}/podman` and reloaded, with a grep confirmation after the edit.
- **crun runtime path**: podman searches a built-in ordered list of absolute paths for the crun OCI runtime, which puts the system `/usr/bin/crun` ahead of the bundle's `/usr/local/bin/crun`. The system crun is too old for podman 5.x.  A user-level `[engine.runtimes] crun = ["/usr/local/bin/crun"]` override points podman at the bundle copy.
- **User-scoped config**: minimal `~/.config/containers/{containers,storage,registries}.conf` to avoid overriding system config files. `mount_program` points to `/usr/bin/fuse-overlayfs` (distro setuid fusermount3).
- **Verification steps**: exact-version assertions with hard failure for both podman and podman-compose, not warnings.
- **CI cost / event gating**:
  - pull_request: the podman lane is gated like the other integration jobs (see #67), so PRs are not slowed by integration tests.
  - push-to-main: one representative leg (5.3.2/1.3.0) runs.
  - schedule and workflow_dispatch: the full four-leg sweep runs.

## Review fixes applied

- Static job name so skipped-PR rendering works correctly
- Restored `etc/` extraction from the static bundle to `/etc`
- `curl -fsSL --retry 3` with per-version pinned sha256
- `exit 1` on version mismatch with exact comparison
- podman-compose version verified with `podman-compose version --short`
- AppArmor profile edit confirmed with `grep -q`
- Runner pinned to `ubuntu-24.04`
- Removed stale `#1341` commit trailers (nonexistent issue)
- Restored comments explaining the /bin symlink guard, `# See #67`, and the ansible-galaxy retry loop
- Replaced the stale `(#155)` reference in the AppArmor comment
- Restored the unit test protecting the PR gate, now asserting the gate directly
- Matrix-builder job gated like integration-podman and pinned to ubuntu-24.04
- Non-podman integration jobs excluded from schedule runs so the weekly trigger only runs the podman sweep
- integration-core scoped to push/workflow_dispatch so it does not join the weekly run

## Known limitations

- `mgoltzsche/podman-static` tops out at v5.8.4, so this lane cannot test podman 6.x.
- Installing the bundle's `etc/containers/` replaces the runner's `registries.conf` with an older format; the lane is rootless, so `~/.config/containers/registries.conf` wins for our tests.

## Verification

- [x] All 4 matrix legs pass on a workflow_dispatch run against the current head (see link below)
- [x] `podman info`, `podman pull`, and `podman run` succeed for every version
- [x] Existing `validate_definitions`, `unit-tests`, and yamllint checks pass

## CI runs

- Full four-leg workflow_dispatch run (current head): https://github.com/hexmode/Canasta-CLI/actions/runs/30564625713
- Fork runs: https://github.com/hexmode/Canasta-CLI/actions

Supersedes/closes #1337
Refs: #1334

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -123,131 +123,100 @@ jobs:
   # Docker is taken away rather than left idle: a hardcoded `docker` call has
   # to fail here instead of quietly reaching a live daemon.
   #
-  # Both runtimes are pinned to exact versions (see the job's env below):
-  # podman 4.9.3 with podman-compose 1.3.0, the pairing the Podman runtime
-  # fixes were verified against. podman-compose 1.6.0 is not covered; the
-  # pull step's reliance on podman-compose skipping buildable services by
-  # default is confirmed for 1.3.0 only.
-  #
-  # A pin the lane cannot enforce is worse than none, because the workflow
-  # then claims a version it is not running. Runner image
-  # ubuntu24/20260726.254 ships its own podman ahead of /usr/bin on PATH:
-  # apt installed 4.9.3 and `podman --version` reported 5.8.4, which failed
-  # the 4.14.1 release on a commit that touched only VERSION and
-  # RELEASE_NOTES (#1333). So the pinned binary is forced onto PATH and the
-  # result confirmed before any test runs. Whether 5.8.4 breaks create at
-  # all is #1334.
+  # A matrix tests three podman versions (4.9.5, 5.3.2, 5.8.4) and two
+  # podman-compose versions (1.3.0, 1.6.0). The full podman binary with its
+  # runtime helpers (crun, conmon, netavark, fuse-overlayfs) is downloaded as
+  # a self-contained static bundle from mgoltzsche/podman-static.  No apt
+  # podman package is needed — the bundle includes everything (#1337).
   integration-podman:
-    name: "Integration: Podman"
-    if: github.event_name != 'pull_request'  # See #67
+    name: "Integration: Podman ${{ matrix.podman_version }}"
+    if: github.event_name != 'pull_request'
     runs-on: ubuntu-latest
     timeout-minutes: 40
-    # Job-scoped so the install and the PATH check cannot disagree.
-    env:
-      PODMAN_VERSION: "4.9.3+ds1-1ubuntu0.2"
-      PODMAN_COMPOSE_VERSION: "1.3.0"
+    strategy:
+      fail-fast: false
+      matrix:
+        podman_version: ["4.9.5", "5.3.2", "5.8.4"]
+        podman_compose_version: ["1.3.0"]
+        include:
+          - podman_version: "5.8.4"
+            podman_compose_version: "1.6.0"
     steps:
       - uses: actions/checkout@v6
       - uses: actions/setup-python@v6
         with:
           python-version: "3.12"
 
-      - name: Install Podman and podman-compose
+      - name: Install system dependencies
         run: |
           sudo apt-get update
           sudo apt-get install -y git-crypt uidmap
 
-          if ! sudo apt-get install -y "podman=$PODMAN_VERSION"; then
-            echo "::error::podman $PODMAN_VERSION is no longer installable" \
-                 "from the archive. Available:"
-            apt-cache madison podman || true
-            echo "Pick a 4.x version from that list, update PODMAN_VERSION," \
-                 "and note the change — this lane's results are only" \
-                 "comparable across runs at a fixed runtime (#1333)."
-            exit 1
-          fi
-
-          # Onto PATH, not into the repo venv: the CLI shells out to
-          # `podman-compose` by name, so it has to resolve like an
-          # operator's install would.
-          python -m pip install --user "podman-compose==$PODMAN_COMPOSE_VERSION"
-          echo "$HOME/.local/bin" >> "$GITHUB_PATH"
-
-      # Installing the pinned package is not enough on its own. Runner image
-      # ubuntu24/20260726.254 ships its own podman ahead of /usr/bin on PATH,
-      # so apt installed 4.9.3 while `podman --version` reported 5.8.4 — and
-      # the lane failed the 4.14.1 release on a commit that touched only
-      # VERSION and RELEASE_NOTES (#1333). Move any such binary aside so the
-      # pinned one is what runs, then confirm it.
-      - name: Make the pinned podman the one on PATH
+      - name: Download podman ${{ matrix.podman_version }} static bundle
         run: |
-          apt_podman="$(dpkg -L podman | grep -E '^/usr/bin/podman$' | head -1)"
-          [ -n "$apt_podman" ] || {
-            echo "::error::the podman package does not own /usr/bin/podman"
-            exit 1
-          }
+          url="https://github.com/mgoltzsche/podman-static/releases/download/v${{ matrix.podman_version }}/podman-linux-amd64.tar.gz"
+          curl -sL "$url" -o /tmp/podman-static.tar.gz
+          sudo tar -xzf /tmp/podman-static.tar.gz -C /usr/local --strip-components=2
+          rm -f /tmp/podman-static.tar.gz
+          echo "/usr/local/bin" >> "$GITHUB_PATH"
 
-          # Compare canonical paths, not strings: /bin is a symlink to
-          # /usr/bin, so `type -a -P` lists the pinned binary twice and a
-          # string compare moves it aside as if it were a foreign one —
-          # which is what broke this step on its first run. The Docker
-          # removal below hits the same symlink and handles it by checking
-          # `-e` after each move.
-          pinned="$(readlink -f "$apt_podman")"
+      # Runner image ubuntu24 ships podman ahead of /usr/local/bin on PATH.
+      # Move those aside so the downloaded one is what runs.
+      - name: Shadow other podman binaries on PATH
+        run: |
+          downloaded="$(readlink -f /usr/local/bin/podman)"
           for p in $(type -a -P podman || true); do
             [ -e "$p" ] || continue
-            if [ "$(readlink -f "$p")" != "$pinned" ]; then
+            if [ "$(readlink -f "$p")" != "$downloaded" ]; then
               echo "Shadowing $p ($("$p" --version 2>/dev/null || echo unknown))"
               sudo mv "$p" "$p.shadowed"
             fi
           done
           hash -r
 
-          resolved="$(command -v podman || true)"
-          [ -n "$resolved" ] || {
-            echo "::error::no podman on PATH after shadowing — the pinned" \
-                 "binary at $apt_podman was moved aside by mistake (#1333)."
-            exit 1
-          }
-          version="$(podman --version | awk '{print $3}')"
-          echo "podman under test: $resolved ($version)"
-          echo "podman-compose:    $(podman-compose --version 2>&1 | tail -1)"
-
-          # The pin is the mechanism; this only proves it took effect. It can
-          # still fail if the image overwrites /usr/bin/podman itself, which
-          # leaves nothing to shadow.
-          case "$version" in
-            "${PODMAN_VERSION%%+*}") ;;
-            *)
-              echo "::error::expected podman ${PODMAN_VERSION%%+*} but" \
-                   "$version resolved from $resolved. The pinned package is" \
-                   "installed, so something on PATH is still shadowing it" \
-                   "(#1333). Whether this version works at all is #1334 —" \
-                   "do not read this as a product regression."
-              exit 1
-              ;;
-          esac
-
-      # Rootless Podman prerequisites the create pre-flight enforces: without
-      # lingering, systemd reaps the containers when the session ends; with
-      # the unprivileged port floor above 80, the pre-flight refuses to run
-      # at all (it cannot see the instance's HTTP_PORT that early).
-      - name: Satisfy rootless Podman prerequisites
+      - name: Configure rootless Podman
         run: |
+          mkdir -p ~/.config/containers
+          cat > ~/.config/containers/registries.conf <<'EOF'
+          unqualified-search-registries = ["docker.io"]
+          EOF
+          cat > ~/.config/containers/storage.conf <<'EOF'
+          [storage]
+          driver = "overlay"
+          [storage.options.overlay]
+          mount_program = "/usr/bin/fuse-overlayfs"
+          EOF
+          # Rootless Podman prerequisites
           sudo loginctl enable-linger "$(id -un)"
           sudo sysctl -w net.ipv4.ip_unprivileged_port_start=80
-          grep -q "^$(id -un):" /etc/subuid || sudo usermod \
-            --add-subuids 100000-165535 --add-subgids 100000-165535 \
-            "$(id -un)"
+          grep -q "^$(id -un):" /etc/subuid 2>/dev/null || \
+            echo "$(id -un):100000:65536" | sudo tee -a /etc/subuid
+          grep -q "^$(id -un):" /etc/subgid 2>/dev/null || \
+            echo "$(id -un):100000:65536" | sudo tee -a /etc/subgid
           echo "XDG_RUNTIME_DIR=/run/user/$(id -u)" >> "$GITHUB_ENV"
+
+      - name: Install podman-compose ${{ matrix.podman_compose_version }}
+        run: |
+          python -m pip install --user \
+            "podman-compose==${{ matrix.podman_compose_version }}"
+          echo "$HOME/.local/bin" >> "$GITHUB_PATH"
+
+      - name: Verify podman version
+        run: |
+          resolved="$(command -v podman || true)"
+          version="$(podman --version | awk '{print $3}')"
+          echo "podman:  $resolved ($version)"
+          echo "target:  ${{ matrix.podman_version }}"
+          target="${{ matrix.podman_version }}"
+          if [ "${version%.*}" != "${target%.*}" ]; then
+            echo "::warning::expected ${{ matrix.podman_version }}" \
+                 "but got $version from $resolved"
+          fi
 
       - name: Remove Docker so only Podman is reachable
         run: |
           sudo systemctl stop docker.socket docker.service || true
           sudo systemctl mask docker.socket docker.service || true
-          # /bin is a symlink to /usr/bin, so `type -a -P` reports the
-          # same binary twice; the second path is already gone once the
-          # first is moved.
           for d in $(type -a -P docker || true); do
             if [ -e "$d" ]; then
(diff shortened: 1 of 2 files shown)


cicalese (commented):
## Review

The direction is right — testing 5.x is exactly what #1334 needs and what the old lane deliberately couldn't do, and a static bundle is a better mechanism than apt pinning. But the PR also undoes the one invariant #1333 established, and the extraction is subtly wrong.

### Blocking

**1. The version assertion is downgraded from a hard failure to a warning — this regresses #1333, which the PR claims to close.**

```bash
if [ "${version%.*}" != "${target%.*}" ]; then
  echo "::warning::expected ${{ matrix.podman_version }}" \
       "but got $version from $resolved"
fi
```

#1333's conclusion was explicit: *"A pin that the lane cannot enforce is worse than no pin, because the workflow then documents a version it is not running."* With a warning, a leg named `Integration: Podman 4.9.5` goes green having tested 5.8.4, and nobody reads warnings on a passing job. (Carried over from #1337 rather than introduced here, but it needs fixing before either lands.) Two further problems: the comparison is major.minor only (`4.9.5` vs `4.9.3` passes) even though the exact version is known from the download; and the removed `[ -n "$resolved" ]` guard means a `podman` accidentally moved aside by the shadow loop yields an empty `$version` and still only warns. Please restore `exit 1` and compare exactly — the download makes the expected version unambiguous, so any mismatch is a broken shadow step, not a judgment call.

`Closes #1333` is also already-closed (by #1335/#1336), so that line reads as re-fixing something this PR actually reverts.

**2. The job name uses a matrix expression — and this PR's own check list shows why the workflow avoids that.**

`.github/workflows/tests.yml` documents the convention 40 lines below, on `integration-features`:

> The job name is static (no matrix expression) so GitHub renders the leg suffix correctly even when the job is skipped on pull_request events

`integration-podman` has the same `if: github.event_name != 'pull_request'`, and `gh pr checks 1340` on this PR reports literally:

```
Integration: Podman ${{ matrix.podman_version }}	skipping	0
```

Also: two legs resolve to the identical name `Integration: Podman 5.8.4` (5.8.4/1.3.0 and 5.8.4/1.6.0), and renaming the check from `Integration: Podman` breaks any branch-protection required-check entry. Suggest a static `name: "Integration: Podman"` plus an early step echoing the leg's two versions.

**3. `--strip-components=3` misplaces the bundle's `etc/containers` tree.** Verified against the real v5.8.4 tarball — the archive root holds both `usr/` and `etc/`:

```
podman-linux-amd64/usr/local/bin/podman        → /usr/local/bin/podman        ✓
podman-linux-amd64/etc/containers/policy.json  → /usr/local/policy.json       ✗
podman-linux-amd64/etc/containers/seccomp.json → /usr/local/seccomp.json      ✗
podman-linux-amd64/etc/containers/{containers,storage,registries}.conf → /usr/local/*.conf  ✗
```

Consequences worth being explicit about, because they invalidate the job comment's claim that *"No apt podman package is needed — the bundle includes everything"*:

- `policy.json` and `seccomp.json` land where podman never looks. The lane only works because the runner image's preinstalled `containers-common` supplies `/etc/containers/policy.json`. On an image without podman preinstalled, `podman pull` fails outright.
- The hand-written `~/.config/containers/containers.conf` re-types `cgroup_manager = "cgroupfs"` and `events_logger = "file"` — which are *exactly and only* what the bundle's own `etc/containers/containers.conf` sets. It is compensating for the dropped file.
- Three other new steps depend on the distro stack the lane claims not to need: `/usr/bin/crun` (the reason for the runc override), `/usr/bin/fuse-overlayfs`, and `/etc/apparmor.d/podman`. Net effect: the lane requires apt podman to be installed while hiding its binary.

Commit 05db9c19 did this correctly (`-C /etc --strip-components=2 podman-linux-amd64/etc/`), and 4c3ac009 removed it while adding the unrelated AppArmor fix, with no reason recorded in the message or the diff. If installing the bundle's `/etc/containers` broke something, that is the single most valuable comment to put in this lane; if it was collateral, restore it. (With it restored you would still need the `mount_program` override or `chmod u+s /usr/local/bin/fusermount3` — the bundle's `fusermount3` is archived 0755, so that comment is correct.)

**4. The download is unverified.** `curl -sL` with no `-f`, no retries, no integrity check, feeding `sudo tar` into `/usr/local`:

- Without `-f`, a bad tag yields a 404 body written to the tarball and `curl` exits 0; the failure surfaces as an opaque `tar: not in gzip format`.
- Upstream publishes a detached signature (`podman-linux-amd64.tar.gz.asc`) for every release and documents verification with key `0CCF102C4F95D89E583FF1D4F8B5AF50344BB503`. This lane swapped a signed apt archive for an unauthenticated third-party binary extracted as root. At minimum use `curl -fsSL --retry 3` plus a per-version pinned sha256; the gpg path is better if flaky keyservers are tolerable.

**5. podman-compose's version is never verified, and the description overstates the shadowing.** The matrix now varies podman-compose, but nothing asserts which one resolved — the exact gap #1333 described, reproduced in the second dimension. Related: the description says *"pre-installed podman/podman-compose binaries are moved aside"*, but the loop only iterates `type -a -P podman`; `podman-compose` is never shadowed (it happens to work because `$HOME/.local/bin` is prepended via `GITHUB_PATH`).

**6. #1334 is still open and 5.8.4 is now half the matrix.** If the 35-minute silent `canasta create` hang reproduces, `main` gets two permanently red legs — correctly, but nobody will be able to tell a new regression from the known one. If the fork runs show 5.8.4 passing end to end, that is significant evidence for #1334 and belongs on that issue; either way this PR should say which of the two it is.

### Non-blocking

- **Issue references.** `Fixes #1341` — no such issue exists (highest is #1339); four commit trailers also say `#1341`. `Closes #1333` is already closed. `Closes #1337` closes a PR — this PR's own predecessor — not an issue. So this PR references no open issue, although #1334 is the one it bears on. The comment ref `(#155)` resolves in this repo to an unrelated merged PR, "Add some output messages"; the intended source is `containers/podman#24642` / the podman-static README. Per repo convention refs belong in commits/PRs rather than in comments anyway.
- **Switching upstreams dropped podman 6.x coverage.** #1337 targeted 4.9.5 / 5.8.4 / 5.8.5 / 6.0.2; this PR targets 4.9.5 / 5.3.2 / 5.8.4 because `mgoltzsche/podman-static` tops out at v5.8.4. Changing binary source was the right call — `podman-container-tools/podman` publishes only `podman-remote-static-linux_amd64.tar.gz` for Linux, which is the remote client, not an engine — but the consequence is that the lane cannot test podman 6.x at all, and 6.0.x is what Ubuntu 26.04 users will be on. That is the same shape as #1333: the runtime moves underneath us and CI has nothing to say about it. Worth recording as a known limitation in the job comment, with whatever the plan is for 6.x (distro packages on a newer runner, or building the engine). 5.3.2 as a 5.x mid-point is a good pick given #1334's note that 5.4.2 works by hand.
- **The AppArmor step fails open.** A non-matching profile leaves `sed` exiting 0, `apparmor_parser` errors go to `/dev/null || true`, and a missing profile file skips silently — three ways to no-op into the `reexec: Permission denied` class of failure with no signal. Worth a `grep -q 'local/bin' /etc/apparmor.d/podman` confirmation after the edit.
- **`runs-on: ubuntu-latest`** while the new steps hardcode runner-image specifics (`/usr/bin/crun`, `/usr/bin/fuse-overlayfs`, the exact `profile podman /usr/bin/podman ` line). The description says ubuntu-24.04; pinning `runs-on: ubuntu-24.04` would make that claim enforceable and is consistent with this lane's stated philosophy.
- **Runtime fidelity.** `runtime = "runc"` + `cgroupfs` + `events_logger = "file"` means the lane exercises a combination no operator runs (Ubuntu podman defaults to crun with the systemd cgroup manager). `[engine.runtimes] crun = ["/usr/local/bin/crun"]`, or shadowing `/usr/bin/crun` the same way podman is shadowed, keeps the bundle coherent. The comment's "hardcoded path" framing is also imprecise — podman searches an ordered list of absolute paths with `/usr/bin` first; nothing is hardcoded to a single path.
- **The `storage.conf` comment is inverted.** "to avoid overriding system config files" — a user `storage.conf` fully replaces `/etc/containers/storage.conf`, unlike `containers.conf`, which merges.
- **Comments deleted while the code they explain stays.** The `/bin`-is-a-symlink note is gone from the Docker-removal loop but the `[ -e "$d" ]` guard it justifies remains — that check now looks removable. Same for `# See #67` on the `if:`, and for the ansible-galaxy retry rationale, which is untouched code unrelated to this PR. Most of the trimming is a welcome tightening; these three losses are not.
- **CI cost.** 4 legs × up to 40 min on every push to `main`, on a lane that exists inside #67's wall-clock budget. Consider one canonical leg per push with the version sweep on `schedule` / `workflow_dispatch`.
- **The shadow step's premise no longer holds.** `GITHUB_PATH` prepends `/usr/local/bin`, so it already wins; the comment "ships podman ahead of `/usr/local/bin` on PATH" is stale. Keep the loop as defense, but note that the download *overwrites* any `/usr/local/bin/podman`, leaving nothing to shadow in that case.
- **subuid/subgid** via `tee -a` instead of `usermod --add-subuids`: the range is equivalent (100000–165535 = 65536 ids) and the guard is right; it does skip usermod's validation, so a stray manual edit will not be caught.

### Verifiability

This lane is skipped on `pull_request`, so nothing here can be validated before merge — the only evidence is the fork runs, and the description offers checkboxes rather than run URLs. Given that #1333 and #1334 both originated in unverifiable claims about this exact lane, the four run links (and the podman/podman-compose versions each actually resolved) are cheap insurance before this lands on `main`.

cicalese (commented):
## Follow-up on ad3b0462

Thanks — nearly all of the first round is handled. I checked the three pinned checksums by downloading the actual release files, and all three match. The `/etc` extraction is correct now, and the hard-failing version check is back.

Two new problems came in with this commit, though, and each one fails every leg.

### 1. podman lands in the wrong directory

```bash
sudo tar -xzf /tmp/podman-static.tar.gz \
  -C /usr/local --strip-components=2 podman-linux-amd64/usr/
```

Stripping two levels removes `podman-linux-amd64/` and `usr/`, which leaves `local/bin/podman`. Extracted into `/usr/local`, that becomes `/usr/local/local/bin/podman`. I ran both forms against the real v5.8.4 archive:

```
/usr/local/local/bin/podman   ← as written
/usr/local/bin/podman         ← with -C /usr --strip-components=2
```

So podman never reaches the PATH, and the job stops at "Verify podman version". This is the same off-by-one that ab5681e3 fixed, coming back from the other direction.

Change `-C /usr/local` to `-C /usr` on that line. The `/etc` line right below it is correct as is — please leave that one alone.

### 2. The podman-compose check cannot pass

```bash
version="$(podman-compose --version | awk '{print $NF}')"
if [ "$version" != "${{ matrix.podman_compose_version }}" ]; then ... exit 1
```

`podman-compose --version` prints two lines, not one — its own version, then the podman version underneath. Both 1.3.0 and 1.6.0 do this from the same code:

```python
print("podman-compose version", __version__)
await compose.podman.run(["--version"], "", [])
```

`awk '{print $NF}'` picks up the last word of both lines, so the variable ends up holding `1.3.0` and `5.8.4` together and never equals `1.3.0`. The old code's `| tail -1` was working around this.

Use `podman-compose version --short`, which prints the version number by itself. Note that `--short` belongs to the `version` subcommand, so `--version --short` will not work.

### One more thing

The description still says all four legs pass. That was true of the earlier commit, but it cannot hold for this one. Since both #1333 and #1334 started from claims about this lane that nobody could check, please add links to the four runs.

### Smaller items still open from the first round

- The `(#155)` reference in the AppArmor comment points at an unrelated merged PR in this repository.
- Three comments were dropped while the code they explain stayed: the note about `/bin` being a symlink, which is why the Docker loop checks `[ -e "$d" ]`; `# See #67` on the `if:` line; and the reason for the ansible-galaxy retry loop, which this PR does not otherwise touch.
- Now that the bundle's own `/etc/containers/containers.conf` is installed, it is worth rechecking whether the `runtime = "runc"` override is still needed.
- Four legs at up to 40 minutes each, on every push to `main`. A single leg per push with the rest on a schedule would cost a lot less.
- The description says the runner puts its podman ahead of `/usr/local/bin` on the PATH. Entries added through `GITHUB_PATH` go to the front, so that is no longer the case.
- The `/etc` extraction replaces the runner's `/etc/containers/registries.conf` with the bundle's older-format one. Rootless runs are unaffected because the user's own file wins, but anything running podman as root would read the older file.
- #1337 is still open and this description no longer mentions it, so the duplicate is now unlinked.

cicalese (commented):
## Follow-up on 3447d44b

Both blockers from the last round are fixed, and the logs confirm it end to end:

```
podman-compose: /home/runner/.local/bin/podman-compose (1.6.0)   target: 1.6.0
podman:         /usr/local/bin/podman (5.8.4)                    target: 5.8.4
```

All four legs ran the full suite and passed. The `(#155)` reference is gone, the three dropped comments are back, and pointing crun at the bundle copy is a better fix than forcing runc. Nice work.

**This is also the first real evidence on #1334.** Both 5.8.4 legs completed `create` and the rest of the suite with no hang. Please add those run links to that issue — it is the only place that evidence exists right now.

### Please put the pull-request gate back

`3fbd2e04` makes this lane run on pull requests, and `3447d44b` rewrites the test that protected the old behavior:

```python
-    def test_podman_lane_is_gated_like_the_other_integration_jobs(self):
-            "the Podman lane must carry the same event gate as the other "
-            "integration jobs (see #67), so it is not accidentally the "
-            "only one running on pull requests"
+    def test_podman_lane_runs_on_pull_requests(self):
```

We do not want to add roughly fifteen minutes to every pull request. That is the exact cost #67 removed, and it lands on every change, including one-line documentation edits. My earlier comment asked for links to your runs, not for a change in when the lane runs.

The change also cuts coverage where we had it: push-to-main now runs a single leg, so main gets less than it does today.

What I would suggest instead, keeping your dynamic matrix as is:

- pull requests: lane does not run, same gate as the other integration jobs
- push to main: the single representative leg you already defined
- schedule and manual runs: the full four-leg sweep

That gives the version coverage you are after without slowing anyone down, and manual runs are the escape hatch #67 called for when a change needs a pre-merge check.

Please also restore the original test. Beyond the policy, the replacement is weak: `podman_if is None or podman_if != "github.event_name != 'pull_request'"` passes for any string that is not that exact one, so a typo in the gate would sail through.

If you want to make the case for running integration jobs on pull requests, that is a fair discussion to have on #67 as its own change.

### Two smaller things

The `schedule:` entry sits at the top of the file, so the whole suite now runs weekly, not just the podman sweep. That includes the Kubernetes, GitOps, and remote-host lanes, unattended. If a weekly podman sweep is the goal, the other jobs need to be excluded from schedule runs.

The new matrix-builder job has no event gate and runs on `ubuntu-latest`, while the job it feeds is pinned to `ubuntu-24.04`.

The branch is also four commits behind main.

### Not your bug, for the record

The scary red blocks in the logs are Ansible printing a full failure for tasks that are then ignored, all during teardown after the test has already passed. The Docker lane on main has more of them than your lane does. Nothing here is masking a failed test. I am filing three separate issues for the underlying causes, including one this lane is the only thing that surfaces: on a podman-only host, `create` probes `docker info` and prints a failure before falling back correctly.

cicalese:
Correction to my first review: I said `Fixes #1341` pointed at an issue that did not exist. It does exist — a closed issue about `PHP_POST_MAX_SIZE` interpolation, fixed by #1344. Both the REST and GraphQL lookups returned 404 when I checked, and I took that at face value instead of digging further.

Dropping the reference was still the right move, since that issue has nothing to do with the Podman lane, but my reason for it was wrong. Sorry for sending you after it on a bad premise.

cicalese (commented):
## Follow-up on a7bd8bda

The gate is back on both podman jobs, the original test is restored, the matrix builder is pinned, the other lanes are scoped off the weekly run, and the branch is rebased. That covers everything from the last round. Unit tests, lint, helm lint and the smoke test all pass, and the podman legs correctly show as skipped here.

Two things left.

### integration-core still runs weekly

The commit says non-podman integration jobs are excluded from schedule runs so only the podman sweep runs weekly. Six jobs were moved to `push || workflow_dispatch`. `integration-core` was left on `!= 'pull_request'`, which is true for a schedule run, so Core Lifecycle — the longest of them — joins every Sunday run.

This is really an interlock, not an oversight. The restored test says:

```python
assert (jobs["integration-podman"].get("if")
        == jobs["integration-core"].get("if"))
```

The podman gate has to allow `schedule` for the weekly sweep to happen at all, and the test forces core's gate to match podman's, so core cannot be scoped without turning the test red.

I would fix the test rather than the workflow. What #67 actually asks for is that the lane does not run on pull requests, so assert that directly instead of comparing the string to another job's gate. Then core can be scoped like its six siblings and the podman lane keeps its schedule run.

Worth noting the test's own message has drifted too: "the same event gate as the other integration jobs" is now true of exactly one of the seven.

### The run link says the wrong trigger

The body cites run 30558574353 as the latest full upstream run on workflow_dispatch or schedule. That run's event was `pull_request` — it is the run from the window when the lane was temporarily ungated, on head 3447d44b.

The evidence itself is good, and I read those logs: four legs, five of five tests each. But given how much of this lane's history has been about which version and which trigger actually ran, the label should match the run. A `workflow_dispatch` run on the current head would fix the label and confirm the restored gate still leaves the sweep reachable — worth doing once before merge.

Those same run links belong on #1334, which is still open and still says create hangs on 5.8.4.

Issue #1296, opened 2026-07-29
Podman has no readiness gate at all: create races composer and fails intermittently


cicalese:
`canasta create` is a coin flip on Podman. The post-merge run of #1292 on `main` failed with:

```
PHP Fatal error: Uncaught Error: Failed opening required
'/var/www/mediawiki/w/vendor/composer/../aws/aws-sdk-php/src/functions.php'
```

`install.php` ran while composer was still writing `vendor/`. That is exactly the race the web health gate exists to prevent — the comment above it in `roles/orchestrator/tasks/start.yml` names this symptom precisely:

> /run-all.sh in the canasta image runs `composer update` synchronously before starting apache; install.php and any other `docker compose exec web php …` task that runs immediately after `up -d` will race composer and crash with "vendor/composer/... bootstrap.php: No such file or directory" if vendor is mid-write.

In that job the gate was **skipped 7 times out of 7**. On Podman it is never entered, so there is no readiness gate at all.

## How this happened

#1283 fixed the gate hanging for its full 60 x 10s on Podman 4.9.3, where `.State.Health` exists but `.Status` renders empty. The fix treats an empty render as "no health signal to wait on" and skips the wait.

That reading is right for a custom image with no `HEALTHCHECK`. It is wrong for Podman, where the healthcheck *is* defined and simply never runs. The risk was noted in #1281 and left open:

> Worth deciding separately whether skipping the wait is safe enough on Podman: the comment above the gate explains it exists to stop `up -d` racing composer. Skipping restores that race on runtimes with no health signal.

It was not decided, and the race has now materialized. It is intermittent, not deterministic — in the same job `lifecycle` and `upgrade-rebuilds-buildable` both created instances successfully and only `upgrade` lost the race. Three passes, one failure.

## Why the status is empty

The signal is not missing, it is never triggered. The image declares a healthcheck:

```
{"Test":["CMD-SHELL","wget -q --method=HEAD localhost/server-status"],
 "Interval":60000000000,"Timeout":10000000000,"StartPeriod":300000000000}
```

Podman inherits that definition but schedules healthchecks with systemd transient timers, which podman-compose does not set up. So the check is configured and never executed, and `.State.Health.Status` stays empty indefinitely.

## Suggested fix

[...]

PR #1359, opened 2026-07-31
fix(start): drive Podman's readiness gate from the play

### LINKED ISSUE #1356: Podman has no readiness gate: the image's healthcheck is invisible to podman because the config is OCI-format
On Podman, `canasta create` and `canasta start` proceed without waiting for web to be ready. Both arms of the health gate in `start.yml` skip, so nothing holds the play back while composer is still writing `vendor/`, and `install.php` can run against a half-written tree:

```
PHP Warning:  require(/var/www/mediawiki/w/vendor/composer/../aws/aws-sdk-php/src/functions.php):
  Failed to open stream: No such file or directory in
  /var/www/mediawiki/w/vendor/composer/autoload_real.php on line 45
PHP Fatal error: Uncaught Error: Failed opening required '.../aws/aws-sdk-php/src/functions.php'
```

That is [run 30564941145](https://github.com/CanastaWiki/Canasta-CLI/actions/runs/30564941145) on `main` today — the Podman lane, 4 passed 1 failed, `upgrade-rebuilds-buildable`.

This is #1296 again. It was closed by #1297, which added `podman healthcheck run` polling to restore the gate. That fix cannot engage, for a reason one level below where #1296 looked.

## Why

The published image declares a healthcheck, but its config is OCI:

```
$ curl ghcr.io/v2/canastawiki/canasta/manifests/latest   (amd64 child)
mediaType:         application/vnd.oci.image.manifest.v1+json
config mediaType:  application/vnd.oci.image.config.v1+json
Healthcheck:       {"Test": ["CMD-SHELL", "wget -q --method=HEAD localhost/server-status"],
                    "Interval": 60s, "Timeout": 10s, "StartPeriod": 300s}
```

`Healthcheck` is a Docker schema2 field. BuildKit writes it into an OCI config anyway and Docker reads it back; podman ignores it there. Pulled with podman 5.4.2:

```
$ podman image inspect -f '{{.ManifestType}}' ghcr.io/canastawiki/canasta:latest
application/vnd.oci.image.manifest.v1+json
$ podman image inspect -f '{{if .HealthCheck}}{{.HealthCheck.Test}}{{else}}NONE{{end}}' ...
NONE
```

So under Podman the container has no healthcheck at all. Both gate arms then skip:

- `Wait for web container to report healthy` — skipped, `.State.Health.Status` renders empty
- `Wait for web to pass its healthcheck (Podman)` — skipped, because `podman healthcheck run` returns `no defined healthcheck`, and that string is exactly the escape hatch #1297 added for custom images that declare no healthcheck

The escape hatch fires for the **stock** image. Docker meanwhile takes the first arm normally:

```
FAILED - RETRYING: [localhost]: orchestrator : Wait for web container to report healthy (59 retries left).
FAILED - RETRYING: [localhost]: orchestrator : Wait for web container to report healthy (58 retries left).
ok: [localhost]
```

## The stated mechanism is also wrong

`start.yml:201` and #1297 both say podman "never runs it under podman-compose: checks are scheduled with systemd transient timers that compose does not create." That is not what happens. On a Debian 13 host with podman 5.4.2 rootless, systemd cgroups and journald, an image built `--format docker` with a `HEALTHCHECK`, brought up by podman-compose 1.3.0:

```
container inherited:          [CMD-SHELL true]
t+0s health=[healthy]
systemd user health timers:   0
```

Podman inherited the check and evaluated it, with no timer involved. The difference between that image and ours is the manifest format, not systemd.

## Verified outside CI

Reproduced on a hand-test host — Debian 13, podman 5.4.2 rootless, podman-compose 1.3.0, systemd cgroups, journald — so this is not a GitHub-runner artifact. `canasta create` on podman, then inspecting the web container:

```
health_status=[]
container_healthcheck=NONE
on_demand: Error: container ... has no defined healthcheck
```

The same host, same image, same CLI, with Docker as the runtime instead:

```
health=[healthy]  hc=[CMD-SHELL wget -q --method=HEAD localhost/server-status]
wget_rc=0         server-status: HTTP/1.1 200 OK
status in mods-enabled: 2
```

Docker sees the check, runs it, and it passes. Podman sees nothing to run.

**The fix's mechanism is confirmed on the same run.** `db` and `varnish` both report health under podman:

```
abpod_db_1      | Up 9 minutes (healthy)
abpod_varnish_1 | Up 9 minutes (healthy)
abpod_web_1     | (no health)
```

The difference is where the check is declared. `db` and `varnish` declare theirs in the compose file, so podman-compose passes `--health-cmd` and podman honors it. `web` relies on the image, which podman cannot read. That is the whole bug in one line of output.

## Fix

Declare the check on the `web` service in `roles/orchestrator/files/compose/docker-compose.yml`, mirroring the image's own:

```yaml
    healthcheck:
      test: ["CMD-SHELL", "wget -q --method=HEAD localhost/server-status"]
      interval: 60s
      timeout: 10s
      start_period: 300s
```

podman-compose passes that to podman as `--health-cmd` regardless of image format — the same reason `db` and `elasticsearch` already report health on Podman today. Docker is unaffected: a compose-level check overrides an identical image-level one.

Publishing the image in Docker format instead would also fix it, but that lives in the Canasta repo and changes the artifact for every consumer; the compose-level declaration is contained here.

Note this does **not** re-open #484. That was `depends_on: condition: service_healthy` asking compose's resolver to gate startup, which is a separate thing and stays as `service_started`.

## Before shipping: one unresolved measurement

Turning the gate on activates a 60 x 10s wait that has never run on Podman. That must be validated first, and one observation says it may not pass.

On the podman host above, after a successful create, `wget -q --method=HEAD localhost/server-status` inside the web container returned 4 (no connection), and after restarting the stack it returned 8 (server error response) continuously for 312s without ever succeeding — while the identical command under Docker returns 0 and `HTTP/1.1 200 OK`.

That single observation was not reproduced: later attempts on that host were blocked by leftover volumes from an earlier failed create, not by anything in the product. So it is a warning, not a finding.

If it holds, declaring the check would make `create` spend ten minutes failing a check that cannot pass — the shape of #1334, manufactured rather than fixed. Confirm `/server-status` answers 200 inside a Podman-run web container before implementing, and if it does not, find out why the same image serves it under Docker and not under Podman; that difference is then the more interesting bug.

cicalese:
Podman renders an empty health status for the stock image, so the Docker-shaped wait skips and `create` races composer — `install.php` crashing on a half-written `vendor/` is the failure the gate exists to prevent (#1296, and again on `main` in [run 30564941145](https://github.com/CanastaWiki/Canasta-CLI/actions/runs/30564941145)).

Two independent causes, both outside this repo: the published image is OCI-format and podman ignores `Healthcheck` in an OCI config, and podman runs healthchecks from systemd transient timers that compose never creates — the static builds used in CI have no systemd at all.

So the runtime cannot be asked to run the check. Make the request from the play instead, over `exec`, which depends on nothing but the container being up.

## Why not the compose-level healthcheck

That was #1357, and it is reverted in #1358. Declaring the healthcheck does make a status render — and wherever podman cannot run the check, that status stays `starting` until the 60 x 10s budget is spent and `create` fails. It turned an intermittent race into a guaranteed failure on every Podman create.

The lesson is in the design here: this wait is **best-effort**. `failed_when: false` with `until` retries the full budget and then continues. A custom image that never serves `/server-status`, or a runtime quirk nobody has seen yet, costs a warning rather than a create. The gate did not run on Podman at all before, so continuing is never worse than the old behavior.

`127.0.0.1`, not `localhost`: `mediawiki.conf` serves `/server-status` only when `REMOTE_ADDR` is exactly `'127.0.0.1'`, and `localhost` resolves to `::1` first under rootless podman, which falls through to MediaWiki as a 404 (CanastaWiki/CanastaBase#226).

## Verified in both configurations

The two environments whose difference caused #1357 to pass review and then break `main`.

**Static bundle, no systemd** — [dispatch on this branch](https://github.com/CanastaWiki/Canasta-CLI/actions/runs/30597149829), all four podman legs plus Docker and K8s green. From the 4.9.5 and 5.8.4/1.6.0 legs:

```
TASK [orchestrator : Wait for web to answer on Podman]
FAILED - RETRYING: ... (29 retries left).
FAILED - RETRYING: ... (28 retries left).
FAILED - RETRYING: ... (27 retries left).
ok: [localhost]
TASK [orchestrator : Report that web never answered on Podman]
skipping: [localhost]

=== Results: 5 passed, 0 failed, 0 skipped ===
```

Zero warnings emitted across every start in both legs, so the gate genuinely engaged rather than falling back to warn-and-continue. Worth stating explicitly: because the wait is best-effort, a green lane alone would not have proved this — the logs had to be read.

**Distro podman 5.4.2 with systemd, Debian 13** — two retries then `ok`, `create_rc=0` in 151s, `/server-status` reachable.

## Guards

The Podman unit tests keep every original intent and gain one that would have caught #1357 at commit time:

```python
def test_it_cannot_fail_the_create(self):
    assert wait.get("failed_when") is False, (
        "…Making this wait fatal is what broke every Podman create once before")
```

Also pinned: the request must address `127.0.0.1` and must not contain `localhost`; the budget must stay bounded; a skipped gate must be reported rather than silent.

`make test-unit` (2259 passed), `make lint`, `validate_definitions` pass.

Closes #1356

### DIFF
--- a/roles/orchestrator/tasks/start.yml
+++ b/roles/orchestrator/tasks/start.yml
@@ -198,52 +198,59 @@
         - _start_web_cid.stdout | trim != ''
         - _start_web_has_health.stdout | default('') | trim != ''
 
-    # Podman inherits the image's healthcheck but never runs it under
-    # podman-compose: it schedules checks with systemd transient timers
-    # that compose does not create, so .State.Health.Status stays empty
-    # and the wait above is skipped for every container. Skipping it is
-    # right for a custom image that declares no healthcheck, and wrong
-    # here — it restores the composer race the gate exists to prevent.
-    # `podman healthcheck run` executes the configured check on demand,
-    # so poll that instead of a status nothing updates.
-    - name: Probe whether web defines a healthcheck (Podman)
+    # Podman renders an empty status for the stock image, so the wait
+    # above skips and the composer race comes back. Two reasons, both
+    # outside our control: the published image is OCI-format and podman
+    # ignores Healthcheck in an OCI config, and podman runs healthchecks
+    # from systemd transient timers that compose never creates — the
+    # static builds used in CI have no systemd at all.
+    #
+    # So ask the container the question the healthcheck would ask,
+    # rather than asking the runtime for an answer it cannot produce.
+    # This depends on nothing but `exec`.
+    #
+    # 127.0.0.1, not localhost: mediawiki.conf serves /server-status
+    # only when REMOTE_ADDR is exactly '127.0.0.1', and localhost
+    # resolves to ::1 first under rootless podman, which falls through
+    # to MediaWiki as a 404.
+    #
+    # Best-effort on purpose. A custom image that never serves
+    # /server-status must not turn a missing signal into a failed
+    # create — the gate did not run on Podman at all before this, so
+    # warning and continuing is never worse than the old behavior,
+    # while a wait that cannot pass would be far worse. 30 × 10s
+    # matches the image healthcheck's 5 min --start-period.
+    - name: Wait for web to answer on Podman
       ansible.builtin.command:
         argv:
           - "{{ _inspect_cmd }}"
-          - healthcheck
-          - run
+          - exec
           - "{{ _start_web_cid.stdout | trim }}"
-      register: _start_web_hc_probe
+          - wget
+          - "-q"
+          - "--method=HEAD"
+          - "127.0.0.1/server-status"
+      register: _start_web_podman_ready
       changed_when: false
       failed_when: false
+      retries: 30
+      delay: 10
+      until: _start_web_podman_ready.rc == 0
       when:
         - _start_web_cid.stdout | trim != ''
         - (_start_web_has_health.stdout | default('') | trim) == ''
         - _inspect_cmd | default('docker') is search('podman')
 
-    # Same budget as the Docker path: 60 × 10s, over the healthcheck's
-    # 5 min start period plus a slow first composer install. A container
-    # with no healthcheck at all must skip rather than spend that budget
-    # failing, and podman distinguishes the two cases in its message.
-    - name: Wait for web to pass its healthcheck (Podman)
-      ansible.builtin.command:
-        argv:
-          - "{{ _inspect_cmd }}"
-          - healthcheck
-          - run
-          - "{{ _start_web_cid.stdout | trim }}"
-      register: _start_web_podman_health
-      changed_when: false
-      retries: 60
-      delay: 10
-      until: _start_web_podman_health.rc == 0
-      when:
-        - _start_web_cid.stdout | trim != ''
-        - (_start_web_has_health.stdout | default('') | trim) == ''
-        - _inspect_cmd | default('docker') is search('podman')
-        - >-
-          'no defined healthcheck'
-          not in (_start_web_hc_probe.stderr | default(''))
+    - name: Report that web never answered on Podman
+      ansible.builtin.debug:
+        msg: >-
+          web did not answer /server-status within 5 minutes, so the
+          readiness gate is being skipped. A custom image that does not
+          serve /server-status is expected here. On the stock image this
+          means web is still starting, and the MediaWiki installer may
+          race composer — check the container logs if create then fails
+          on a missing vendor/ file.
+      when: (_start_web_podman_ready.rc | default(0)) != 0
 
     # Register CAPI + auto-enroll the bouncer when CrowdSec is on. Shared
     # with the K8s start path and create. crowdsec_autoenroll is false only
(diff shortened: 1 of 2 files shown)


roles/orchestrator/tasks/start.yml
```diff
@@ -198,52 +198,59 @@
...
-    # and the wait above is skipped for every container. Skipping it is
-    # right for a custom image that declares no healthcheck, and wrong
-    # here — it restores the composer race the gate exists to prevent.
-    # `podman healthcheck run` executes the configured check on demand,
-    # so poll that instead of a status nothing updates.
-    - name: Probe whether web defines a healthcheck (Podman)
+    # Podman renders an empty status for the stock image, so the wait
+    # above skips and the composer race comes back. Two reasons, both
+    # outside our control: the published image is OCI-format and podman
+    # ignores Healthcheck in an OCI config, and podman runs healthchecks
```
hexmode, line 204: https://github.com/podman-container-tools/podman/issues/25454 Support Docker’s non-OCI health check in OCI images?

PR #1371, opened 2026-08-01
fix(upgrade): detect the runtime of an instance that records none

### LINKED ISSUE #1370: docker still expected in podman instance during upgrade
```
mah@flynn:~/home-files/org$ canasta list
Canasta ID      Host              Instance Path
  Orchestrator  Status
  Wiki ID       Wiki URL
──────────────────────────────────────────────────────
nichework       nichework.com     /home/mah/canasta/nichework
  COMPOSE       STOPPED
  nichework     wiki.nichework.com/
mah@flynn:~/home-files/org$ canasta doctor -i nichework
Canasta Dependency Check (nichework.com)
====================================================

Core (required):
  Python 3:        OK (Python 3.13.5)
  Docker:          MISSING
  Docker Compose:  MISSING
  Docker daemon:   NOT RUNNING
  Rootless socket: unix:///run/user/1001/podman/podman.sock (canasta create will auto-set --docker-host to this)
  Privileged ports: OK (ip_unprivileged_port_start=80)
  Podman:          5.4.2 (OK)

Kubernetes (optional):
  kubectl:         not installed
  Helm:            not installed
  k3s:             not installed
  Cluster:         not reachable
  Argo CD:         not found

GitOps (optional):
  git:             OK (git version 2.47.3)
  git-crypt:       OK
  sops:            not installed (needed on the gitops host for encrypted K8s secrets; 'canasta install sops')
  age:             not installed (operator key tool for encrypted K8s secrets; 'canasta install sops')

Scheduled backups (optional):
  crontab:         OK
  canasta on host: not installed (scheduled backups run 'canasta backup create' on the host; install with 'canasta install canasta -H <host>' or canasta-docker)

System:
  Memory:          15 GB
  Disk (/ avail):  124G
  www-data group:  NOT A MEMBER — Canasta containers write files as www-data; add yourself with: sudo usermod -aG www-data $USER (then log out and back in, or start a new shell, for the change to take effect)
  canasta group:   NOT A MEMBER — 'canasta upgrade' self-updates by writing the install dir; without membership the git fetch is skipped and the CLI silently stays on the old version. Add with: sudo usermod -aG canasta $USER (then log out and back in)

Instance consistency (nichework):
  OK (profiles, running services, and search backend agree)
mah@flynn:~/home-files/org$ canasta upgrade
Pulling canasta-ansible image (ghcr.io/canastawiki/canasta-ansible:4.14.1)...
[...]
Pulling Canasta container images...
Error: pull Compose images failed (rc=127): /bin/sh: 1: docker: not found

### PR DESCRIPTION
**Pull request by cicalese**
Fixes #1370.

## Cause

`composeCommand` / `inspectCommand` only became registry fields in #1224, so any instance registered before that carries neither. `_upgrade_single.yml` reads the runtime off the registry record and falls back to the play-scope `docker compose` default when it is absent — the registry's silence is indistinguishable from a recorded Docker instance. On a Podman-only host the whole upgrade then runs Docker commands, and `pull` is the first step that cannot tolerate it:

```
Pulling Canasta container images...
Error: pull Compose images failed (rc=127): /bin/sh: 1: docker: not found
```

`canasta doctor` on the same host reports Docker MISSING and Podman OK, which is what makes the failure confusing.

#1272 fixed the read-the-runtime-from-the-registry half of this. The remaining hole is the instances that have nothing recorded.

The same gap makes such an instance read as STOPPED in `canasta list`: the direct-command fast path resolves it to `["docker", "compose"]` too, and for a remote instance the `.env` fallback never applies.

## Fix

- `roles/common/tasks/detect_runtime.yml` (new): probes `docker info` then `podman info` on the instance's own host and, when Docker is absent and Podman is present, sets the runtime facts and writes them back to the registry. The write-back is what repairs `list` / `start` / `exec` as well, so the probe is not repeated per command.
- `playbooks/_upgrade_single.yml`: includes it after the connection switch, only for compose instances whose registry entry records no runtime.
- `roles/common/library/canasta_registry.py`: new `state: set_runtime`. `state: present` rebuilds the record from module params, so backfilling one field through it would silently drop `devMode`, `buildFrom`, `dockerHost` and the rest.

Only the Podman case is persisted. A working Docker host already matches the default, and a Docker daemon that merely happens to be down during the run is not evidence worth writing to the registry — so Docker hosts re-probe on each upgrade, at the cost of one command.

## Testing

New `tests/unit/test_upgrade_detects_legacy_runtime.py` (14 tests) covers the include and its guards, the probe's tolerance of a missing binary (`docker info` leaves no `rc` at all when Docker is not installed), the write-back, and `set_runtime`'s field preservation and idempotence.

`make test-unit` (2320 passed), `make validate` and `make lint` all pass.

### DIFF
--- a/playbooks/_upgrade_single.yml
+++ b/playbooks/_upgrade_single.yml
@@ -27,6 +27,16 @@
 - name: Switch connection to instance host
   ansible.builtin.include_tasks: "{{ canasta_root }}/roles/common/tasks/switch_connection.yml"
 
+# An instance registered before the runtime became a registry field has
+# nothing for the facts above to read, so they fall back to Docker and
+# the whole upgrade runs `docker compose` — rc=127 on a Podman-only
+# host. Silence in the registry is not evidence of Docker; ask the host.
+- name: Detect the runtime for an instance that has none recorded
+  ansible.builtin.include_tasks: "{{ canasta_root }}/roles/common/tasks/detect_runtime.yml"
+  when:
+    - instance_orchestrator == 'compose'
+    - _inst.value.composeCommand is not defined
+
 - name: Run upgrade role (migrations + refresh + restart)
   ansible.builtin.include_role:
     name: upgrade
--- a/roles/common/library/canasta_registry.py
+++ b/roles/common/library/canasta_registry.py
@@ -28,6 +28,7 @@ options:
       - query
       - query_all
       - query_by_path
+      - set_runtime
       - set_setting
       - get_setting
     default: query
@@ -154,7 +155,7 @@ def run_module():
     module_args = dict(
         state=dict(type="str", default="query",
                    choices=["present", "absent", "query", "query_all",
-                            "query_by_path",
+                            "query_by_path", "set_runtime",
                             "set_setting", "get_setting"]),
         setting_key=dict(type="str", required=False),
         setting_value=dict(type="str", required=False),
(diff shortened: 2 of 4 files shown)


hexmode:
And here, I just discovered the inspectCommand setting and was about to update this...

Issue #1382, opened 2026-08-01
canasta start returns before web is ready: the Compose fast path has no readiness gate


cicalese:
Found in a hands-on e2e of `release/4.15.0` on a Docker Compose instance.

`canasta start` returns as soon as `docker compose up -d` exits, without waiting for the web container to become ready. The Ansible start path has a readiness gate; the `direct_commands` fast path that actually serves Compose instances has none.

## Measured

```
12:09:33  canasta stop
12:09:38  canasta start RETURNED
12:09:40  web = starting
12:09:45  web = starting
12:09:51  web = starting
12:09:57  web = healthy
```

`start` returned 19 seconds before the wiki could serve. On a first start after an image change, or any start where `composer update` re-runs, that window is minutes rather than seconds.

## Cause

`direct_commands/lifecycle.py::cmd_start` falls back to Ansible only for Kubernetes and for instances with sidecars. Every other Compose instance takes the fast path:

```python
_helpers._sync_compose_profiles(inst)
rc = _helpers._run_compose(inst_id, inst, ["up", "-d"])
return rc
```

There is no equivalent of the gate in `roles/orchestrator/tasks/start.yml`, which waits `retries: 60, delay: 10` for the container to report `healthy`, and on Podman probes `/server-status` directly.

## Why it matters

The gate's own comment in `start.yml` states the hazard it exists to prevent:

> `/run-all.sh` in the canasta image runs `composer update` synchronously before starting apache; `install.php` and any other `docker compose exec web php …` task that runs immediately after `up -d` will race composer and crash with "vendor/composer/... bootstrap.php: No such file or directory" if vendor is mid-write.

`canasta start && canasta maintenance script …` is exactly that sequence, and on the fast path nothing prevents it. `create` is unaffected because it drives the Ansible start internally, which is why this does not show up in a normal create test.

## Scope note

This is not a regression in 4.15.0 — the fast path has been gateless since the `direct_commands` split (`769ee9791`, #495, May 2026). It is worth recording now because it also bounds the reach of #1377: the improved readiness-timeout diagnosis added there is in the Ansible gate, so it never runs for `canasta start` on a plain Compose instance.

## Suggested direction

Give the fast path the same contract as the Ansible one — poll the container's health status after `up -d`, with the same empty-status/Podman `exec` fallback — or have `cmd_start` return `FALLBACK` when the instance's image declares a healthcheck, so the gate runs where it matters.

cicalese:
[...]

PR #1395, opened 2026-08-01
fix(status): join the ps command before interpolating it into the remote shell

### LINKED ISSUE #1394: podman-compose: running start on an already-running instance fails with container name conflicts
When `canasta start` is run on an instance that is already running under Podman, container creation fails because the container names already exist:

```
$ ./canasta-docker start -i nichework
nichework_db_1
Error: creating container storage: the container name "nichework_db_1" is already in use by 786c436060...
You have to remove that container to be able to reuse that name: that name is already in use, or use --replace to instruct Podman to do so.
```

Docker Compose handles this gracefully (it sees the containers exist and treats them as already running), but podman-compose fails with name conflicts when the containers already exist.

**Expected behavior:** `canasta start` on an already-running instance should be a no-op (or recreate the containers), not fail with container name conflicts.

**Environment:**
- Podman 5.4.2
- podman-compose (version TBD)
- Canasta CLI native install via canasta-docker

### LINKED ISSUE #1396: fix: canasta status crashes on remote host with '[podman-compose,: command not found'
Running `canasta status -i <id>` on an instance whose host is not localhost produces a shell error:

```
$ ./canasta-docker status -i nichework
Instance:     nichework
Host:         nichework.com
Orchestrator: COMPOSE
Status:       RUNNING
bash: line 1: [podman-compose,: command not found

Containers:
  (query failed or no containers)
```

Root cause: `direct_commands/info.py:497` interpolates a Python list (`ps_cmd`) into an SSH command string via `%s`. Python's `str()` on a list produces the bracketed repr like `['podman-compose', 'ps']`, which the shell then tries to execute as `[podman-compose,:` — a command that does not exist.

The localhost path correctly passes the list to `subprocess.run()` which accepts lists. The remote path needs `" ".join()` to convert the list to a shell command string, matching the pattern already used in `_helpers.py:_check_running_compose`.

hexmode:
Fixes #1396

`canasta status` against a remote host crashes:

```
[podman-compose,: command not found
```

`ps_cmd` is built as a list (`compose_cmd + ["ps"]`), and the remote branch interpolated it into the SSH command with `%s`, stringifying the list itself:

```
cd /srv/wiki && ['podman-compose', 'ps']
```

Only the remote branch was affected. The local branch passes the list straight to `subprocess`, where a list is exactly what is wanted — which is why every local test passed while this was broken.

## Changes

**`direct_commands/info.py`** — join before interpolating:

```python
ps_str = " ".join(ps_cmd)
rc, out = _helpers._ssh_run(
    host, "cd %s && %s" % (_helpers._shell_quote(path), ps_str),
)
```

**Tests**, covering this and adjacent behavior that had none:

- `test_status_remote_ps_command.py` — the remote command is a string, the compose command survives intact, the path is still shell-quoted, and the *local* branch still passes a list (so "fixing" both branches the same way is caught). Reverting the join fails three of the five.
- `test_start_noop_when_running.py` — a running instance skips compose and returns 0, a stopped one still starts, and the readiness gate runs after a real start. That last one fails if `_wait_web_ready` is ever dropped from `cmd_start` again, which is how #1382 would silently regress.
- `test_check_running_is_running_only.py` — `check_running.yml` stays running-only. It is shared with `upgrade` (which reads it into `_was_running` to decide whether to restart afterward) and `delete`, so widening it to `ps -a` would make `upgrade` start an instance that was deliberately stopped.

## Scope

This PR started as the `start` no-op for #1394. That landed separately in #1393, which is already on `main` and closed #1394, so the no-op and the `start.yml` change are gone from here — `direct_commands/lifecycle.py` and `roles/orchestrator/tasks/start.yml` are byte-identical to `main`.

Also removed along the way, with the reasoning in the comments below: a `podman-compose ps` fallback whose premise did not hold (the `canasta-ansible` image ships neither podman nor podman-compose), a `_start_or_noop` extraction with one caller that hid the readiness wait behind its name, and a `cmd_stop` early return keyed on the web container alone — which would have reported "already stopped" while `db`, `caddy` or `varnish` were still running.

## Verified

```
make lint        yamllint + ruff + ansible-lint + validate, all clean
                 ansible-lint: 0 failures in 347 files, profile "production"
make validate    87 commands, 69 playbooks, 161 examples
make test-unit   2413 passed
```


cicalese:
Rebased onto `main` now that #1405 has landed. The conflict in `cmd_start` was real and I resolved it to compose both behaviors rather than take either side:

```python
    rc = _helpers._start_or_noop(inst_id, inst)
    if rc != 0:
        _helpers._dump_compose_failure(inst)
        return rc
    return _helpers._wait_web_ready(inst_id, inst)
```

Taking this branch as-is would have dropped `_wait_web_ready` (the readiness gate from #1386/#1382, which ships in 4.15.0) and `_dump_compose_failure`. All 13 tests in `test_start_readiness_gate.py` pass against the result. The test stub conflict needed both `_check_running_compose` and `_wait_web_ready` mocked, since `cmd_start` now calls one through `_start_or_noop` and the other directly.

Also restored the two explanatory comments on the `FALLBACK` branches, which the branch had dropped.

`make lint` clean, `make test-unit` 2395 passed.

One open question before merge — `check_running.yml` is changed from `ps` to `ps -a`, and that task is shared:

- `roles/upgrade/tasks/main.yml:30` reads `_container_running` into `_was_running`, which decides whether `upgrade` restarts the instance afterward
- `roles/orchestrator/tasks/delete_cleanup_files.yml:30` includes it too

With `-a`, a **stopped** container satisfies `stdout | length > 0`, so `_container_running` becomes true for a stopped instance and `upgrade` would start one that was deliberately down. If the start path needs "a container exists" rather than "a container is running", it may be worth a separate check rather than widening the shared one.

cicalese:
Decision on the `ps -a` change: **reverted, in both places.** Pushed as `d5ff36ffb`.

`check_running.yml` is now byte-identical to `main` again, and the compose fallbacks in `start.yml` use `ps -q` rather than `ps -a -q`.

## Why

Widening the probe broke it in two directions.

**Other callers.** `check_running.yml` is shared. `roles/upgrade/tasks/main.yml:30` reads `_container_running` into `_was_running`, which decides whether the upgrade restarts the instance when it finishes; `roles/orchestrator/tasks/delete_cleanup_files.yml:30` includes it too. With `-a`, a stopped container satisfies the non-empty test, so an instance that was deliberately stopped looks running — and `upgrade` would start it back up. That is a behavior change to two commands from a PR about `start`.

**The start path itself.** #1394 is specifically about running `canasta start` on an instance that is *already running*. Skipping when a container merely *exists* would make `canasta start` a no-op on a stopped instance and leave it down — the containers are still present after a crash, a reboot without a restart policy, or a `docker stop`, so the skip would fire exactly when the operator wants a start. (`canasta stop` runs `down --remove-orphans`, which removes them, so the stop/start path was never the case that exposed this.)

So running-only is the right semantics for both, and the shared task did not need to change at all.

## Known gap

This leaves one case unhandled: containers that exist but are **stopped** under podman-compose. `up -d` can still hit the name conflict there, and this PR no longer skips it. That is a narrower problem than the one #1394 reports, and the fix is to start or recreate the existing containers rather than to skip — worth its own issue rather than widening a shared readiness probe to cover it. Happy to file that if you agree.

## Test

`tests/unit/test_check_running_is_running_only.py` asserts the shared probe does not list all containers, that the start fallbacks agree, and that both other callers still read it — with the reasoning in the failure messages. The `ps -a` in the compose failure dump is deliberately untouched and has its own assertion: a diagnostic *should* show stopped containers.

`make lint` clean, `make test-unit` 2401 passed.

cicalese:
Added the missing coverage (`926f508bf`), and one scoping correction.

## #1394 is already fixed on `main`, by #1393

Both entry points skip a running instance on `main` today — the inline `_check_running_compose` pre-check in `cmd_start`, and `Skip start when already running` at `start.yml:77`. I wrote tests for that behavior and **they pass unmodified against `main`**, so this branch is not what fixes #1394. I closed #1394 against #1393 rather than leaving it attached here.

That makes this PR's real contribution the `canasta status` fix (#1396), the compose-`ps` fallbacks, and the refactor of the pre-check into `_start_or_noop`.

## Tests

`tests/unit/test_status_remote_ps_command.py` — the #1396 fix had no coverage, and the bug is remote-only, which is why every local test passed while it was broken. Reverting the join makes three of the five fail. It also pins the local branch to the list form, so "fixing" both branches the same way would be caught.

`tests/unit/test_start_compose_ps_fallback.py` — the fallbacks were the genuinely new logic here and had no coverage: they run only when the native probe came back empty, are running-only, cannot fail the start, and OR into `_container_running` rather than overwriting it.

`tests/unit/test_start_noop_when_running.py` — a running instance skips compose and still returns 0, a stopped one still starts, and the readiness gate still runs after a real start. As noted these pass on `main` too, so they are regression cover for the `_start_or_noop` refactor rather than proof of a new fix. The last one is the one that matters most: it fails if `_wait_web_ready` is ever dropped from `cmd_start` again.

`make lint` clean, `make validate` clean, `make test-unit` 2417 passed.

## Follow-up filed

#1408 — containers that exist but are **stopped** under podman-compose still hit the name conflict on `up -d`. The skip deliberately does not fire there, since a stopped instance should start; the remedy is to start or recreate the existing containers. Out of scope here.

cicalese:
Stripped the `_start_or_noop` refactor. In doing so I found a second change that had not been discussed, and removed that too — details below, because it is the more important of the two.

## The refactor

Three reasons it did not earn its place:

- **One caller.** `_start_or_noop` was called only from `cmd_start`, so the extraction bought indirection and no reuse.
- **The name understated it.** Its own docstring: *"Run `up -d` if the instance is not already running, then wait for the web container to become ready."* The readiness wait is the #1382 fix, and burying it behind a name about the no-op is exactly where it should not be.
- **It broke a helper's contract.** `_dump_compose_failure` documents itself as running *"right after the failed 'up' and before the rollback"* — that is why it captures `ps` and logs when it does. In the refactored `cmd_start` it fired on any non-zero rc, including a readiness timeout where `up -d` had **succeeded** and no rollback follows.

Reverting `cmd_start` to main's inline form loses nothing: all sixteen tests on this branch still pass against it.

## The `cmd_stop` change

Stripping the refactor surfaced this. The branch gave `cmd_stop` the same early return:

```python
if not _helpers._check_running_compose(path, host, docker_host, compose_cmd):
    print("Instance '%s' is already stopped." % inst_id)
    return 0
```

`_check_running_compose` filters on `label=com.docker.compose.service=web`. It answers "is the **web** container running", not "is the instance running".

So an instance whose web container is down while `db`, `caddy` or `varnish` are still up reports *"already stopped"* and skips `down --remove-orphans` entirely — leaving those containers running. `canasta stop` would report success without stopping the instance. That is a worse failure than the one the start-side no-op avoids, and unlike that one it is not cosmetic.

It also deleted four explanatory comments from `cmd_stop`, including the two that matter most: why `COMPOSE_PROFILES` must be reconciled before `down` (a drifted profile set leaves a running container as an unmanaged orphan), and why `--remove-orphans` is there at all.

Both files are now byte-identical to `main`.

## What is left

```
direct_commands/info.py                            +3/-1
tests/unit/test_status_remote_ps_command.py        +126
tests/unit/test_start_noop_when_running.py         +109
tests/unit/test_check_running_is_running_only.py    +85
```

The `ps_str` fix — your diagnosis, and a real bug on `main`: `canasta status` against a remote Podman host sends the Python list repr and the shell reports `[podman-compose,: command not found`. Reverting the join fails three of the five tests that now cover it.

Plus regression coverage for behavior that had none: that a running instance skips compose and still returns 0, that a stopped one still starts, that the readiness gate runs after a real start (this one fails if `_wait_web_ready` is ever dropped again), and that the shared `check_running.yml` probe stays running-only so `upgrade` cannot restart a deliberately-stopped instance.

`make lint` clean, ansible-lint 0 failures across 347 files, `make validate` clean, `make test-unit` 2413 passed.

The `test_direct_commands.py` stubs went with the refactor — they existed only to keep those tests working under it, and its 344 tests pass without them.

PR #1397, opened 2026-08-01
fix: join ps_cmd list into string before SSH interpolation

### LINKED ISSUE #1396: fix: canasta status crashes on remote host with '[podman-compose,: command not found'

hexmode:
Fixes #1396

direct_commands/info.py:497 passed a Python list (`ps_cmd`) directly into an SSH command string via `%s`. `str()` on a list produces the bracketed repr like `['podman-compose', 'ps']`, which the shell then tries to execute as `[podman-compose,:` — `command not found`.

The localhost path (line 488) correctly passes the list to `subprocess.run()`. The remote path now joins into a string first, matching the pattern already used in `_helpers.py:_check_running_compose`.

### DIFF
--- a/canasta.py
+++ b/canasta.py
@@ -1515,6 +1515,27 @@ def build_ansible_args(ansible_playbook, command_name, args, data):
         except (OSError, IOError, KeyError):
             pass
 
+    # Infer compose_command from inspect_command or dockerHost. An
+    # instance whose inspectCommand in the registry is 'podman' but
+    # whose composeCommand was never recorded (e.g. from a state=update
+    # that set only the former) would otherwise fall back to 'docker
+    # compose'. A legacy instance that predates the runtime fields
+    # entirely carries only the podman socket in dockerHost.
+    if "compose_command" not in extra_vars:
+        if extra_vars.get("inspect_command") == "podman":
+            extra_vars["compose_command"] = "podman-compose"
+        elif "compose_command" not in extra_vars:
+            try:
+                inst = resolve_instance(
+                    getattr(args, "id", None), required=False
+                ) if os.path.isfile(get_config_file_path()) else None
+                inst = inst or {}
+                docker_host = inst.get("dockerHost") or ""
+                if "podman" in docker_host:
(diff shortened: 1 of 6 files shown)


cicalese:
Superseded by #1395, which now carries this fix.

The `canasta status` bug this PR is named for — `ps_cmd` reaching the remote shell as a Python list repr, producing `[podman-compose,: command not found` — is fixed identically in #1395:

```python
ps_str = " ".join(ps_cmd)
rc, out = _helpers._ssh_run(
    host, "cd %s && %s" % (_helpers._shell_quote(path), ps_str),
)
```

It also now has test coverage there (`tests/unit/test_status_remote_ps_command.py`), which reverting the join demonstrably fails. #1396 has been moved onto #1395 so the issue still closes when that merges.

Comparing the two branches directly, this one adds nothing #1395 lacks. Its only differences are an older `_start_or_noop` that reimplemented the compose invocation — dropping the `_is_localhost`/ssh branch, so `canasta start` would have run locally against a path on the target — and the absence of the tests. #1395 reworked that helper to delegate to `_run_compose`, which keeps the ssh path, `DOCKER_HOST` propagation, the reset retry, and output streaming.

Thanks for the `ps_cmd` diagnosis — it was correct and is shipping, just via the other branch.

Issue #1387, opened 2026-08-01
canasta-docker: upgrade never updates the CLI image, so --dev is a silent no-op


cicalese:
### Summary

In `canasta-docker` mode, `canasta upgrade` never updates the CLI itself, so `--dev` has no effect. The wrapper re-pulls whatever tag is pinned in `$CONFIG_DIR/cli_image_tag`, and nothing in the repo ever rewrites that file except the installer. An operator on an older pinned tag can never move forward, and gets no indication why.

### Mechanism

`self_update_cli()` returns immediately when the script dir has no `.git`:

```python
repo = SCRIPT_DIR
if not os.path.isdir(os.path.join(repo, ".git")):
    return  # Docker install — image pull happened in the wrapper.
```

The wrapper selects the image before that runs:

```bash
PINNED_TAG_FILE="$CONFIG_DIR/cli_image_tag"
DEFAULT_TAG="latest"
if [[ -f "$PINNED_TAG_FILE" ]]; then
    PINNED_TAG="$(tr -d '[:space:]' < "$PINNED_TAG_FILE")"
    [[ -n "$PINNED_TAG" ]] && DEFAULT_TAG="$PINNED_TAG"
fi
```

`cli_image_tag` is written in exactly one place in the repo — `get-canasta.sh`, at install time. So the pin is permanent.

Two comments describe behavior that does not exist:

- `canasta-docker`: "The channel (release vs dev) is pinned by 'canasta upgrade' in $CONFIG_DIR/cli_image_tag"
- `get-canasta.sh`: "A later upgrade rewrites this, so a miss here is harmless"

### Impact

A user pinned to an older tag runs `canasta upgrade --dev`, sees it complete successfully, and is still on the old CLI. Switching their *checkout* to main changes nothing either, since in docker mode the CLI code comes from the image, not the checkout. This makes every already-fixed bug look unfixed.

### Reproduction

With `cli_image_tag` containing an older version, `canasta-docker upgrade --dev` re-pulls that same tag and the CLI version is unchanged.

Current workaround:

```
CANASTA_CLI_IMAGE=ghcr.io/canastawiki/canasta-ansible:main canasta-docker upgrade
```

or `echo main > $CONFIG_DIR/cli_image_tag`.

### Suggested fix

[...]

PR #1404, opened 2026-08-01
fix(docker): report the CLI channel and a wrapper that cannot update

cicalese:
`canasta upgrade` in docker mode did the right thing and said almost nothing about it, which is what made #1387 hard to pin down.

## What was already working

The wrapper does rewrite `$CONFIG_DIR/cli_image_tag` on upgrade — through the `$PINNED_TAG_FILE` variable, which a grep for the literal filename misses. That has been on `main` since #627. Checked against the registry:

- `canasta-ansible:latest` and `:main` are the **same digest**, so `--dev` writing `latest` does track head of main.
- `latest_release_tag()` returns `4.14.1`, and `canasta-ansible:4.14.1`, `:4.14`, `:4.13.0` all resolve.

So the pin moves and both channels point somewhere real. This PR does not change any of that.

## What was missing

**The channel was never reported.** `upgrade` and `upgrade --dev` produced identical output, so an operator who suspected the CLI had not moved had nothing to check. Each branch now names what it resolved to:

```
CLI channel: development (head of main), image ghcr.io/canastawiki/canasta-ansible:latest.
CLI channel: release, image ghcr.io/canastawiki/canasta-ansible:4.14.1.
CLI image: … (pinned by CANASTA_CLI_IMAGE; 'canasta upgrade' will not change it).
```

**A wrapper that cannot update itself was nearly silent.** This is the failure that hides every other one: only a wrapper new enough to read `cli_image_tag` honours the channel written to it, so a wrapper installed before #627 keeps pulling its own baked-in tag and `--dev` appears to do nothing — the exact symptom reported. Four paths lead there and they were reported unevenly:

| Path | Before | After |
| --- | --- | --- |
| sudo refused | one-line warning | ERROR + consequence + fix |
| unwritable, no sudo | one-line warning | ERROR + consequence + fix |
| wrapper fetch failed | **silent** | ERROR + consequence + fix |
| curl not installed | **silent** | ERROR + consequence + fix |

All four now set `WRAPPER_STALE` and print, to stderr:

```
ERROR: cannot write to /usr/local/bin/canasta and sudo is not available.
       This wrapper is out of date and stays out of date.
       Instances below still upgrade, but the CLI itself does not,
       so a channel change ('--dev') may not take effect and fixes
       released since this wrapper was installed stay missing.
       Replace it with:
         sudo curl -fsSL <url> -o <path> && sudo chmod +x <path>
```

The upgrade still proceeds — the instances are upgradeable either way, and failing the run would be worse than a loud warning.

## On the issue's stated mechanism

The issue reports that `cli_image_tag` "is written in exactly one place in the repo — `get-canasta.sh`", and that two comments describe behavior that does not exist. Both readings come from the `$PINNED_TAG_FILE` indirection; the comments are accurate as written, so they are left alone. The reporting gaps above are what remains, and a pre-#627 wrapper on disk explains the reproduction without any repo-side defect.

## Tests

`tests/unit/test_canasta_docker_upgrade_reporting.py` asserts the script text, since the block pulls images and fetches from GitHub before the dry-run harness is reachable: all three channel reports name `$IMAGE`, all four failure paths set `WRAPPER_STALE`, none of them is a bare "Warning", the previously silent paths speak, the notice states both the consequence and the remediation command, and the whole notice goes to stderr.

## Verified

```
bash -n canasta-docker         syntax OK
make lint                      All checks passed
make validate                  87 commands, 69 playbooks, 161 examples
make test-unit                 2339 passed (incl. the 31 existing wrapper tests)
```

Refs #1387

### DIFF
--- a/canasta-docker
+++ b/canasta-docker
@@ -639,10 +639,30 @@ for arg in "$@"; do
             fi
             WRAPPER_URL="https://raw.githubusercontent.com/CanastaWiki/Canasta-CLI/${WRAPPER_REF}/canasta-docker"
 
+            # Say which channel this landed on. Without it, `upgrade`
+            # and `upgrade --dev` are indistinguishable in the output,
+            # and an operator who suspects the CLI did not move has
+            # nothing to check against.
+            if [[ "$ENV_IMAGE_OVERRIDE" == "true" ]]; then
+                echo "CLI image: $IMAGE (pinned by CANASTA_CLI_IMAGE; 'canasta upgrade' will not change it)."
+            elif [[ "$DEV_CHANNEL" == "true" ]]; then
+                echo "CLI channel: development (head of main), image $IMAGE."
+            else
+                echo "CLI channel: release, image $IMAGE."
+            fi
+
             echo "Pulling canasta-ansible image ($IMAGE)..."
             "$CONTAINER_ENGINE" pull "$IMAGE"
 
             # Update the wrapper script itself if the upstream differs.
+            #
+            # A wrapper that cannot replace itself is the one failure that
+            # hides every other fix: the channel written above is only
+            # honoured by a wrapper new enough to read $PINNED_TAG_FILE,
+            # so an old one keeps pulling its own baked-in tag and
+            # `--dev` appears to do nothing. Report it as an error the
+            # operator has to act on, not a passing note.
+            WRAPPER_STALE=false
             if command -v curl &>/dev/null; then
                 TMP_WRAPPER="$(mktemp)"
                 if curl -fsSL "$WRAPPER_URL" -o "$TMP_WRAPPER"; then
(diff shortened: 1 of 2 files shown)


cicalese:
Updated the remediation: the stale-wrapper notice now points at the installer rather than a hand-fetch of the wrapper file.

```
       Reinstall the CLI:
         curl -fsSL https://get.canasta.wiki | bash -s -- --docker
```

Fetching `canasta-docker` by hand replaces that one file and nothing else. The installer places the wrapper, sets the mode, and writes `cli_image_tag` — so the repair restores the whole install rather than just the script that could not replace itself, which matters here because a stale wrapper is precisely one that does not know about the pin file.

The channel the operator asked for is preserved: `--docker --dev` when the upgrade was run with `--dev`, `--docker` otherwise.

Also rebased onto current `main` (now through #1403).

Issue #1415, opened 2026-08-02
Add actionlint and hadolint to CI


cicalese:
Two small static checks with no coverage today. Both are near-silent right now, so adopting them locks in the current state rather than starting a cleanup.

## actionlint — 5 workflows, 2 findings

```
.github/workflows/tests.yml:411  SC2086  Double quote to prevent globbing and word splitting
.github/workflows/tests.yml:431  SC2046  Quote this to prevent word splitting
```

Both are ShellCheck findings inside `run:` blocks, which actionlint surfaces automatically. It also validates workflow schema, expression syntax, and `runs-on` labels — the kind of mistake that otherwise only shows up as a failed run on push.

## hadolint — 4 Dockerfiles, 3 findings in one

Clean: `Dockerfile`, `images/caddy/Dockerfile`, `roles/devmode/templates/Dockerfile.xdebug`.

`tests/integration/remote/Dockerfile`:

```
:7   DL3008  Pin versions in apt-get install
:19  DL3008  Pin versions in apt-get install
:19  DL4006  Set the SHELL option -o pipefail before RUN with a pipe in it
```

`DL4006` is the same missing-`pipefail` class tracked in #1413 for Ansible shell tasks and #1414 for the wrappers.

There is precedent in the org — CanastaBase already carries a `hadolint.yaml`, so the tool and its config conventions are established.

## Suggested

Add both to the Lint job. Fix the three Dockerfile findings and the two quoting findings so each starts green. `DL3008` (pin apt versions) is arguable for a throwaway integration-test image and may deserve an ignore rather than a fix — worth deciding rather than pinning versions that then rot.

Issue #1414, opened 2026-08-02
Add shellcheck to CI for the shell wrappers


cicalese:
Nothing lints the shell. `canasta-docker`, `canasta-native`, and `get-canasta.sh` are 1,444 lines of bash, and `canasta-docker` (857 of them) is the entry point for the install mode the README recommends.

## It is currently clean

Ran `koalaman/shellcheck:stable` over all three, including info level:

```
canasta-native:22:8   SC2155  Declare and assign separately to avoid masking return values
get-canasta.sh:253:5  SC2034  ARCH appears unused
```

Two findings, both cosmetic. So this is not a bug hunt — it is cheap insurance that locks in the current state.

## Why bother, then

The install and upgrade paths are exactly where a silent shell bug is most expensive, because they run before anything else can report a problem. #1387 is a recent example of that surface being hard to reason about: the wrapper's channel selection, pin file, and self-update interact, and the failure mode was invisible output rather than a crash.

`SC2155` in particular masks return values — `local x="$(cmd)"` discards `cmd`'s exit status — which is the same class as the `pipefail` gap tracked in #1413.

There is org precedent: `mywikis/CrawlerProtection` added ShellCheck to CI recently, and `actionlint` already surfaces ShellCheck findings for `run:` blocks in workflows (see the sibling issue), so the tool is partly in use already without covering the wrappers themselves.

## Suggested

Add a shellcheck step to the Lint job covering `canasta-docker`, `canasta-native`, `get-canasta.sh`, and `_sources`-style scripts if any land later. Fix or annotate the two existing findings so it starts green.

Containerised (`koalaman/shellcheck:stable`) avoids adding a system dependency to the venv setup.

PR #1416, opened 2026-08-02
Run ansible-lint, and make shell pipelines fail loudly

### LINKED ISSUE #1413: ansible-lint is pinned in requirements.txt but never run
`requirements.txt` pins `ansible-lint==26.6.0`, so it installs into `.venv` on every setup. Nothing in the `Makefile` or `.github/workflows/` ever calls it.

Ansible is the largest surface in the repo — 293 task and playbook files against 217 Python files — and the only check it gets today is `yamllint --strict`, which validates YAML syntax and style, not Ansible semantics.

## What it finds

1191 findings, but the shape matters more than the total:

| Rule | Count | Worth adopting |
| --- | --- | --- |
| `no-role-prefix` | 908 | **No.** Variable-naming convention; adopting means renaming hundreds of variables across every role for no behavioral change. |
| `command-instead-of-module` | 128 | **Mostly no.** Shelling out to `docker`/`kubectl`/`git` is deliberate here — `exec.yml` is the abstraction for it. A few `git` cases might be worth the module. |
| **`risky-shell-pipe`** | **12** | **Yes.** |
| **`ignore-errors`** | **24** | **Review.** |
| `no-handler` | 8 | Maybe. |
| `command-instead-of-shell` | 6 | Cosmetic. |
| `no-changed-when` | 2 | Yes — cheap. |
| `risky-file-permissions` | 1 | Yes — cheap. |

## `risky-shell-pipe` is the one that matters

A pipeline's exit status is the *last* command's, so `foo | grep bar` reports grep's result and a failing `foo` passes silently unless `pipefail` is set. Twelve shell tasks are exposed:

```
roles/backup/tasks/restore_k8s.yml:411, :475
roles/devmode/tasks/enable.yml:82
roles/gitops/tasks/_detect_gitcrypt_lock.yml:12
roles/gitops/tasks/_heal_dangling_submodules.yml:14
roles/gitops/tasks/_recover_orphan_gitlinks.yml:15
roles/orchestrator/tasks/create_preflight.yml:119, :213, :227, :333
roles/orchestrator/tasks/restore_instance.yml:321
roles/upgrade/tasks/migrations/extract_secret_key.yml:12
```

Several of those are in restore and gitops recovery paths, where a silently-swallowed failure is worst: the command reports success and the operator believes data came back.

This is not theoretical. The same mistake in a shell one-liner during a recent session made a captured exit code read the wrong command's status, and the run looked clean.

## `ignore-errors`

24 tasks use `ignore_errors: true`, which suppresses everything rather than a named condition. The codebase already prefers `failed_when:` in most places — `run_backup.yml` is a good example, tolerating only "already exists" so a real failure still stops the backup. These 24 are where that discipline is missing, starting with:

```
roles/backup/tasks/restore_k8s.yml:286, :300, :313
roles/delete/tasks/main.yml:299, :309
roles/devmode/tasks/enable.yml:76
roles/gitops/tasks/join.yml:194, :338
```

Worth reading rather than bulk-converting — some are deliberate.

## Suggested approach

Mirror the staging in #1409: adopt a narrow rule set first rather than the full default.

```
risky-shell-pipe, ignore-errors, no-changed-when, risky-file-permissions
```

That is ~39 findings, all about correctness. Skip `no-role-prefix` and `command-instead-of-module` explicitly in the config, with the reason recorded, so the next person does not rediscover the question.

Wire it into `make lint` alongside yamllint and ruff, so it runs where the other gates run.

cicalese:
`ansible-lint==26.6.0` is pinned in `requirements.txt` and installs into `.venv` on every setup. Nothing ever ran it. Ansible is the largest surface in the repo — 293 task and playbook files against 217 Python files — and the only check it had was `yamllint --strict`, which validates YAML, not Ansible.

## The rule set

The default reports 1191 findings, which is why turning it on wholesale was never attractive:

| Rule | Count | Enabled |
| --- | --- | --- |
| `no-role-prefix` | 908 | no — renaming hundreds of variables for no behavioral change |
| `command-instead-of-module` | 128 | no — shelling out to docker/kubectl/git is the design; `exec.yml` is the abstraction |
| **`risky-shell-pipe`** | 12 | **yes** |
| `ignore-errors` | 24 | not yet — 24 judgment calls, left in #1413 |
| `no-changed-when` | 5 | **yes** |
| `risky-file-permissions` | 1 | **yes** |

`.ansible-lint` records the skipped families and why, so the question does not have to be rediscovered.

## What `risky-shell-pipe` was hiding

A pipeline's exit status is its **last** command's. Without `pipefail`, `foo | bar` reports `bar`'s result and a failing `foo` passes silently. The worst instance is the Kubernetes DB import:

```yaml
kubectl exec ... cat /currentsnapshot/config/backup/{{ item }} |
kubectl exec -i ... bash -c 'mariadb ...'
```

If the `cat` fails — missing file, pod gone — `mariadb` reads empty input and exits 0. **The restore reports success having imported nothing.** Several of the other eleven are in gitops recovery paths, where the same silence applies.

## `executable: /bin/bash` is not optional here

`pipefail` is not in POSIX until Issue 8, and dash only implemented it in 0.5.12. Debian 13 has 0.5.12 and accepts it — verified on a live host, `set -o pipefail; false | true` gives rc=1 — but Ubuntu 22.04 ships 0.5.11, where the option itself errors. Ansible's `shell` defaults to `/bin/sh`, so every fixed task sets the executable explicitly.

One pre-existing `pipefail` task had the same gap and would have failed on those targets, silently: it carries `failed_when: false`.

## The `kinds` mapping

Everything under `playbooks/` is a task file included from `canasta.yml`, not a play. ansible-lint assumed otherwise and reported 71 `syntax-check` errors of the form "'ansible.builtin.set_fact' is not a valid attribute for a Play" — which were **masking three real `no-changed-when` findings** in `create.yml`'s rescue block. Those are now fixed too.

## Result

```
Passed: 0 failure(s), 0 warning(s) in 347 files processed of 358 encountered.
Last profile that met the validation criteria was 'production'.
```

## Verified

```
make lint                 yamllint + ruff + ansible-lint + validate_definitions, all clean
make validate             87 commands, 69 playbooks, 161 examples
make test-unit            2397 passed
ansible-playbook --syntax-check   ok
```

Every changed task was re-parsed with PyYAML and its `cmd` rendered as the shell would receive it, to confirm the folded-scalar cases got `set -o pipefail;` with the separator and the literal-block cases got their own line.

## Follow-ups

`ignore-errors` (24 sites) stays open on #1413 — each needs a decision about which failure is tolerable, and the codebase already has the better pattern in `run_backup.yml`, which tolerates only "already exists".

Sibling issues from the same survey: #1414 (shellcheck) and #1415 (actionlint, hadolint).

Fixes #1413

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,0 +1,54 @@
+---
+# Narrow by design, in the spirit of ruff.toml. The default rule set
+# reports 1191 findings on this repo, 908 of them `no-role-prefix` — a
+# variable-naming convention whose adoption would mean renaming hundreds
+# of variables across every role for no behavioral change, and 128
+# `command-instead-of-module`, where shelling out to docker/kubectl/git is
+# the deliberate design (that is what exec.yml exists for).
+#
+# What is enabled here is the subset about correctness rather than style.
+# Widen it deliberately, the way #1409 widened ruff, rather than by
+# turning the default set on.
+#
+# See #1413.
+
+# Everything under playbooks/ is a task file included from canasta.yml,
+# not a play. Without this, ansible-lint parses each as a playbook and
+# reports every task keyword as "not a valid attribute for a Play".
+kinds:
+  - tasks: "playbooks/*.yml"
+
+exclude_paths:
+  - .claude/
+  - .venv/
+  - tests/integration/
+
+# risky-shell-pipe        a pipeline reports its LAST command's status, so a
+#                         failing command mid-pipe passes silently unless
+#                         pipefail is set. This hid a restore that could
+#                         report success having imported nothing.
+# no-changed-when         a command task that always reports "changed" makes
+#                         --check and idempotency claims meaningless.
+# risky-file-permissions  a file created without an explicit mode inherits
+#                         whatever umask is in play.
+# ignore-errors           deliberately NOT enabled yet: 24 sites, each needing
+#                         a judgment call about which failure is tolerable.
+#                         Tracked in #1413.
+enable_list:
+  - risky-shell-pipe
+  - no-changed-when
+  - risky-file-permissions
+
+skip_list:
+  # add_host in canasta.yml is inherently a once-per-run action; the rule
+  # only matters under `strategy: free`, which this repo does not use.
+  - run-once
+  - no-role-prefix
+  - command-instead-of-module
+  - command-instead-of-shell
+  - no-handler
+  - name
+  - ignore-errors
+  - jinja
+  - var-naming
+  - yaml
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -33,7 +33,7 @@ exclude_paths:
 #                         whatever umask is in play.
 # ignore-errors           deliberately NOT enabled yet: 24 sites, each needing
 #                         a judgment call about which failure is tolerable.
-#                         Tracked in #1413.
+#                         Tracked in #1417; add it here once they are done.
 enable_list:
   - risky-shell-pipe
   - no-changed-when
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ PIP := $(VENV)/bin/pip
 PYTEST := $(VENV)/bin/pytest
 YAMLLINT := $(VENV)/bin/yamllint
 RUFF := $(VENV)/bin/ruff
+ANSIBLE_LINT := $(VENV)/bin/ansible-lint
 
 # --- Setup -------------------------------------------------------------------
 
@@ -34,6 +35,7 @@ test: test-unit
 lint: venv
 	$(YAMLLINT) --strict meta/ roles/ playbooks/ inventory/ canasta.yml
 	$(RUFF) check .
+	$(ANSIBLE_LINT) --offline roles/ playbooks/ canasta.yml
 	$(PYTHON) scripts/validate_definitions.py
 
 # --- Documentation -----------------------------------------------------------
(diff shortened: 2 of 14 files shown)


cicalese:
Split the `ignore-errors` follow-up into its own issue: #1417. The `.ansible-lint` comment now points there rather than back at #1413, so the note stays accurate once #1413 closes with this PR.

cicalese:
Rebased onto `main` now that #1411 and #1412 have landed. The conflict was in the `lint:` target and was the superseded half of my own change: #1411 moved the ruff selection into `ruff.toml` and reduced the Makefile to `$(RUFF) check .`, while this branch still carried the `--select F --preview` flags from #1410.

Resolved to keep main's form and add only the new line:

```make
lint: venv
	$(YAMLLINT) --strict meta/ roles/ playbooks/ inventory/ canasta.yml
	$(RUFF) check .
	$(ANSIBLE_LINT) --offline roles/ playbooks/ canasta.yml
	$(PYTHON) scripts/validate_definitions.py
```

So `ruff.toml` stays authoritative for ruff, and `.ansible-lint` is authoritative for ansible-lint — the same shape for both tools.

Re-verified against the new base, including that the line-length ceiling #1412 introduced accepts the lines this PR rewrapped:

```
make lint        yamllint + ruff + ansible-lint + validate, all clean
                 ansible-lint: 0 failures in 347 files, profile "production"
make validate    87 commands, 69 playbooks, 161 examples
make test-unit   2397 passed
```

PR #1421, opened 2026-08-02
fix: set inspect_command from dockerHost and preserve fields in registry state:present

### LINKED ISSUE #1419: fix: dockerHost inference in build_ansible_args does not set inspect_command
In `canasta.py:build_ansible_args()`, when inferring the runtime from a legacy instance's `dockerHost` field, only `compose_command` is set to `"podman-compose"` — `inspect_command` is left at the play-scope default `"docker"`.

```python
# Lines 1533-1535
docker_host = inst.get("dockerHost") or ""
if "podman" in docker_host:
    extra_vars["compose_command"] = "podman-compose"
```

This causes Ansible tasks that use `inspect_command` (like `check_running.yml`) to run `docker ps` instead of `podman ps`, which fails on a Podman-only host where the `docker` binary is not installed.

**Impact:** Legacy instances (predating the `composeCommand`/`inspectCommand` registry fields) that have `dockerHost` set pointing at a podman socket will get `compose_command=podman-compose` but `inspect_command=docker`. The Python fast-path (`_resolve_inspect_cmd`) handles this via its own `dockerHost` fallback, but the Ansible path does not.

**Fix:** Add `extra_vars["inspect_command"] = "podman"` alongside the existing `extra_vars["compose_command"] = "podman-compose"` in the dockerHost inference block.

### LINKED ISSUE #1420: fix: _register.yml state:present can silently erase composeCommand or inspectCommand
When `roles/create/tasks/_register.yml` writes to the registry using `state: present`, it uses `default(omit, true)` for both `compose_command` and `inspect_command`:

```yaml
# Lines 24-25
compose_command: "{{ compose_command | default(omit, true) }}"
inspect_command: "{{ inspect_command | default(omit, true) }}"
```

Because `state: present` **replaces** the entire registry record (not a merge), if one of these variables is undefined at registration time, only one field is written — silently **erasing** the other from the registry. If this happened to a Podman instance, it would lose the recorded runtime and fall back to Docker defaults on subsequent operations.

Currently mitigated because `create_preflight.yml` always sets both before `_register.yml` runs, but this is a fragile invariant. Any future code path that calls `state: present` with only one of the two variables defined would silently corrupt the registry.

**Proposed fix:** Change `_register.yml` to use `state: update` instead of `state: present`, which only touches supplied fields and preserves existing ones. This is a safe, merge-style write that cannot erase data.

hexmode:
Fixes #1419
Fixes #1420

### Changes

- **canasta.py**: when inferring compose_command from dockerHost containing 'podman', also set inspect_command to 'podman' so Ansible tasks like check_running.yml use the correct runtime binary
- **canasta_registry.py**: state:present now preserves existing fields that the caller did not supply, preventing a partial write from silently erasing composeCommand or inspectCommand from the registry
- **test_registry_partial_update.py**: updated test to reflect the new preserve-semantics of state:present

### DIFF
--- a/canasta.py
+++ b/canasta.py
@@ -1533,6 +1533,7 @@ def build_ansible_args(ansible_playbook, command_name, args, data):
                 docker_host = inst.get("dockerHost") or ""
                 if "podman" in docker_host:
                     extra_vars["compose_command"] = "podman-compose"
+                    extra_vars["inspect_command"] = "podman"
             except (OSError, IOError, KeyError):
                 pass
 
--- a/roles/common/library/canasta_registry.py
+++ b/roles/common/library/canasta_registry.py
@@ -288,6 +288,13 @@ def run_module():
         })
 
         if inst_id in instances:
+            # Preserve fields the caller did not supply so a caller that
+            # only sets compose_command does not silently erase a prior
+            # inspect_command (and vice versa).
+            for key in ("composeCommand", "inspectCommand", "dockerHost",
+                        "buildFrom", "buildArgs", "registry", "kindCluster"):
+                if key not in new_instance and key in instances[inst_id]:
+                    new_instance[key] = instances[inst_id][key]
             if instances[inst_id] != new_instance:
                 changed = True
                 instances[inst_id] = new_instance
--- a/roles/common/library/canasta_registry.py
+++ b/roles/common/library/canasta_registry.py
@@ -175,6 +175,31 @@ def instance_to_dict(instance):
     return result
 
 
+# Fields instance_to_dict writes unconditionally. Everything else it
+# emits is optional, and `state: present` rebuilds the record from its
+# parameters — so any optional field the caller did not supply would
+# vanish, silently erasing either a runtime fact some other code path
+# discovered (composeCommand, inspectCommand, dockerHost) or an operator
+# setting (devMode).
+#
+# Derived from instance_to_dict rather than listed by hand: a field added
+# there is preserved without anyone having to remember this, which is the
+# failure mode that produced the erase in the first place.
+_ALWAYS_WRITTEN = ("id", "path", "orchestrator")
+
+
+def preserved_on_present():
+    """Optional registry fields, taken from instance_to_dict itself."""
+    probe = instance_to_dict({
+        "id": "probe", "path": "/probe", "orchestrator": "compose",
+        "host": "probe", "devMode": True, "managedCluster": True,
+        "registry": "probe", "kindCluster": "probe", "buildFrom": "/probe",
+        "buildArgs": {"probe": "1"}, "dockerHost": "probe",
+        "composeCommand": "probe", "inspectCommand": "probe",
+    })
+    return tuple(k for k in probe if k not in _ALWAYS_WRITTEN)
+
+
 def run_module():
     module_args = dict(
         state=dict(type="str", default="query",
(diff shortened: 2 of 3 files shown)


cicalese:
Rebased onto the current #1395. That PR has been stripped considerably since this branch forked from it — the `_start_or_noop` extraction, the `cmd_stop` early return, the `podman-compose ps` fallback and the `start.yml` change are all gone, and `lifecycle.py`, `_helpers.py` and `start.yml` are now byte-identical to `main`. Replaying this branch normally would have resurrected all of it, so I replayed only the unique commit with `git rebase --onto`.

What is left here is the actual contribution:

```
canasta.py                                  +1
roles/common/library/canasta_registry.py    +7
tests/unit/test_registry_partial_update.py  +6/-2
```

`make lint` clean, ansible-lint 0 failures across 347 files, `make test-unit` 2413 passed.

Two things worth settling before this merges.

## The class name and docstring now contradict their own test

```python
class TestPresentStillReplaces:
    """create owns the whole record and relies on this."""

    def test_it_preserves_fields_it_was_not_given(self, tmp_dir):
```

The test was renamed from `test_it_drops_fields_it_was_not_given` — correctly, since the behavior changed — but the class is still called `TestPresentStillReplaces` and still asserts in prose that create relies on replace semantics. A reader hitting this later gets the opposite of the truth from the container and the truth from the method.

Worth renaming both, and re-examining the claim itself: if `create` really did rely on `present` clearing stale fields, this change removes that. I could not find a flow where it matters — `delete` removes the record entirely, so a re-create inserts fresh — but the docstring asserts it, so whoever wrote it may have had a case in mind.

## The preserve list is a fixed set of seven keys

```python
for key in ("composeCommand", "inspectCommand", "dockerHost",
            "buildFrom", "buildArgs", "registry", "kindCluster"):
```

Any registry field added later is silently not preserved until someone remembers to extend this tuple — which is the same class of fragility #1420 describes, relocated rather than removed. #1420's own framing ("Any future code path that calls `state: present` with only one of the two variables defined would silently corrupt the registry") applies equally to a future *field*.

Preserving every existing key not supplied would close that, at the cost of making `present` behave like `update`. If the distinction is worth keeping, the list is defensible — but it wants a comment saying it is deliberate and that new fields must be added here, or the next field is a silent regression.

Neither is a blocker; both are cheap now and annoying later.

cicalese:
Recording what the last commit changed, since it merged without a note and it altered this PR's approach rather than just adding to it.

## The preserve list was incomplete

`instance_to_dict` emits thirteen fields: three always, ten conditionally. The hand-written tuple covered seven of the ten. Still erasable by exactly the mechanism this PR was fixing:

- **`host`** — how the controller reaches the instance at all
- **`devMode`** — an operator setting, from `devmode enable`
- **`managedCluster`**

A `state: present` write that omitted any of them dropped it, which is #1420's own description (*"a partial write from silently erasing…"*) applied to a different field.

## So it is derived rather than listed

`preserved_on_present()` computes the set from `instance_to_dict` by probing it with every field populated and subtracting the three that are always written. A field added to `instance_to_dict` is now preserved without anyone having to remember this code exists — and forgetting is what produced the erase in the first place, so removing the chance to forget seemed better than making the list longer.

`TestThePreserveListCannotGoStale` keeps the derivation honest. I verified it catches the regression rather than assuming: restoring the hand-written tuple fails two of its three tests and names the three missed fields in the failure message.

## The test class was renamed

`TestPresentStillReplaces`, docstring *"create owns the whole record and relies on this"*, had come to contradict its own test — which this PR correctly renamed to `test_it_preserves_fields_it_was_not_given`. The container asserted replacement while the method asserted preservation.

Now `TestPresentKeepsWhatItWasNotGiven`, with a docstring separating the two claims: create does supply the whole record, so `present` replacing it is right; what is not right is dropping an optional field the caller left unset, because `composeCommand`/`inspectCommand` come from `detect_runtime`, `dockerHost` from the socket probe, and `devMode` from `devmode enable` — none of which the caller of a given `present` write necessarily knows about.

I could not find a flow that relied on `present` clearing fields (`delete` removes the record entirely, so a re-create inserts fresh), but the old docstring asserted one, so it is worth knowing that claim was retired deliberately rather than overlooked.

Thanks for the `inspect_command` catch — the asymmetry in `build_ansible_args` was real and is now on `main`.

Issue #1435, opened 2026-08-04
Add  target + suggest  as fix for Debian 13's buggy podman-compose 1.3.0


hexmode:
**Background**

Debian 13 (trixie) stable ships `podman-compose` 1.3.0, which has a critical bug: `${VAR:-default}` variable substitution fails when the source and destination variable names match (containers/podman-compose#1105). This breaks Canasta's `docker-compose.yml`, which relies on this pattern extensively.

The fix is upstream in podman-compose >= 1.3.1 / 1.5.0, but no Debian stable update is available yet (Debian bug #1135957). In the meantime, users need a way to install a working version of podman-compose.

**Proposed feature**

Add `canasta install uv` as a new install target that:
1. Installs `uv` (the Astral Python package manager) on the target host via their standalone installer
2. Uses `uv` to install a working version of `podman-compose` outside the system package manager

**Why `uv` specifically**

- `uv` is a fast, standalone Python package manager that doesn't require a system Python virtualenv
- `uv tool install podman-compose` installs the latest working version into an isolated tool environment
- This avoids conflicts with `apt` and allows users to get the fix immediately without waiting for Debian stable updates
- `uv` is becoming the standard Python toolchain and is already used by many projects

**Suggested CLI flow**

```bash
# On the controller (this machine)
canasta install uv -H nichework.com

# Then on the remote host (or via canasta ssh / ansible)
ssh nichework.com
uv tool install podman-compose
# or: uv tool install podman-compose==1.5.0
```

Or as a single command with a new `canasta install` target:
```bash
canasta install podman-compose --via uv -H nichework.com
```

**Integration with `canasta doctor`**

Once `canasta doctor` detects podman-compose 1.3.0 (see #1434), the warning message should suggest:
> "Upgrade with: `canasta install uv -H <host>` then `ssh <host> 'uv tool install podman-compose'`"

**Acceptance criteria**

- [ ] New `canasta install uv` target in the install playbook
- [ ] `canasta doctor` warning for podman-compose 1.3.0 references this install path
- [ ] Documentation update showing the workaround for Debian 13 stable users

**Related issues**

- #1434 — doctor should warn about podman-compose 1.3.0
- containers/podman-compose#1105 — upstream bug
- Debian #1135957 — downstream bug report

PR #1436, opened 2026-08-04
fix(doctor): warn about podman-compose 1.3.0's broken ${VAR:-default} substitution

### LINKED ISSUE #1434: doctor should warn about podman-compose 1.3.0's broken ${VAR:-default} substitution
**Problem**

🔴  1.3.0 (shipped in Debian 13 / trixie stable) has a bug where the shell-default variable-substitution syntax  is **not resolved** when the source variable name matches the destination variable name (e.g. ).

This breaks Canasta's , which uses this pattern extensively in the  section:

With podman-compose 1.3.0, these values are passed **literally** to the containers, causing:
- The web container to wait forever for host 
- PHP warnings: 
- And similar failures for every variable that uses the self-referential default pattern.

**Upstream references**
- podman-compose bug: https://github.com/containers/podman-compose/issues/1105
- Fix PR: https://github.com/containers/podman-compose/pull/1110
- Debian bug report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1135957

**Current state**

The Canasta Dependency Check (localhost)
====================================================

Core (required):
  Python 3:        OK (Python 3.12.13)
  Docker:          OK (Docker version 27.5.1, build 9f9e405)
  Docker Compose:  OK (Docker Compose version v5.1.1)
  Docker daemon:   OK (running)
  Rootless socket: none detected (default /var/run/docker.sock)

Kubernetes (optional):
  kubectl:         OK
  Helm:            OK (v3.21.3+g1ad6e68)
  k3s:             not installed
  Cluster:         not reachable
  Argo CD:         not found

GitOps (optional):
  git:             OK (git version 2.47.3)
  git-crypt:       OK
  sops:            not installed (needed on the gitops host for encrypted K8s secrets; 'canasta install sops')
  age:             not installed (operator key tool for encrypted K8s secrets; 'canasta install sops')

Scheduled backups (optional):
  crontab:         not installed (install cron to use 'canasta backup schedule'; the schedule is a host crontab entry on the instance's host, including the control-plane node for Kubernetes)
  canasta on host: not installed (scheduled backups run 'canasta backup create' on the host; install with 'canasta install canasta -H <host>' or canasta-docker)

System:
  Memory:          15 GB
  Disk (/ avail):  201G
  www-data group:  OK (member)
  canasta group:   NOT A MEMBER — 'canasta upgrade' self-updates by writing the install dir; without membership the git fetch is skipped and the CLI silently stays on the old version. Add with: sudo usermod -aG canasta $USER (then log out and back in)
  Self-update:     OK (install dir writable) command already probes podman version 5.4.2 and warns when podman < 4.4. It does **not** check , so a user on a fully supported host (podman 5.4 + podman-compose 1.3.0) gets a clean bill of health while the stack is actually broken.

**Requested change**

Have Canasta Dependency Check (localhost)
====================================================

Core (required):
  Python 3:        OK (Python 3.12.13)
  Docker:          OK (Docker version 27.5.1, build 9f9e405)
  Docker Compose:  OK (Docker Compose version v5.1.1)
  Docker daemon:   OK (running)
  Rootless socket: none detected (default /var/run/docker.sock)

Kubernetes (optional):
  kubectl:         OK
  Helm:            OK (v3.21.3+g1ad6e68)
  k3s:             not installed
  Cluster:         not reachable
  Argo CD:         not found

GitOps (optional):
  git:             OK (git version 2.47.3)
  git-crypt:       OK
  sops:            not installed (needed on the gitops host for encrypted K8s secrets; 'canasta install sops')
  age:             not installed (operator key tool for encrypted K8s secrets; 'canasta install sops')

Scheduled backups (optional):
  crontab:         not installed (install cron to use 'canasta backup schedule'; the schedule is a host crontab entry on the instance's host, including the control-plane node for Kubernetes)
  canasta on host: not installed (scheduled backups run 'canasta backup create' on the host; install with 'canasta install canasta -H <host>' or canasta-docker)

System:
  Memory:          15 GB
  Disk (/ avail):  201G
  www-data group:  OK (member)
  canasta group:   NOT A MEMBER — 'canasta upgrade' self-updates by writing the install dir; without membership the git fetch is skipped and the CLI silently stays on the old version. Add with: sudo usermod -aG canasta $USER (then log out and back in)
  Self-update:     OK (install dir writable) detect when  1.3.0 is installed and emit a prominent warning telling the user to upgrade. The fix is a single-line patch (delete  in ) or upgrading to podman-compose >= 1.3.1 / 1.5.0.

**Workaround already in place (incomplete)**

 already pins some DB defaults (, ) to  for podman-compose hosts, but it misses all the non-DB variables (, , , , etc.). Pinning every compose default is possible but brittle; warning the user about the buggy runtime is a better UX.

hexmode:
**Problem**

podman-compose 1.3.0 (shipped in Debian 13 / trixie stable) has a bug where the shell-default variable-substitution syntax `${VAR:-default}` is passed **literally** to containers when the source and destination variable names match (e.g. `${MYSQL_HOST:-db}`).

This breaks Canasta's `docker-compose.yml`, which uses this pattern extensively in the `environment:` section.

**What this PR does**

- Adds a `podman-compose version` probe to the doctor script
- Parses the version and emits a prominent warning when 1.3.0 is detected
- Suggests upgrading to >= 1.3.1 or installing via the new `canasta install uv` path (#1435)

**Sample output**

```
Core (required):
  Python 3:        OK (Python 3.13.5)
  Docker:          MISSING
  Docker Compose:  MISSING
  Docker daemon:   NOT RUNNING
  Podman:          5.4.2 (OK)
  Podman Compose:  1.3.0 — WARNING: podman-compose 1.3.0 has a known bug ...
```

**Related issues**

- Fixes #1434
- See also #1435 (install uv target)

### DIFF
--- a/direct_commands/doctor.py
+++ b/direct_commands/doctor.py
@@ -37,7 +37,8 @@ command -v sops >/dev/null 2>&1 && echo OK || echo MISSING; echo "$D"
 command -v age-keygen >/dev/null 2>&1 && echo OK || echo MISSING; echo "$D"
 command -v podman >/dev/null 2>&1 && podman --version 2>/dev/null || echo MISSING; echo "$D"
 command -v sudo >/dev/null 2>&1 && sudo --version 2>/dev/null | head -1 || echo MISSING; echo "$D"
-command -v sudo >/dev/null 2>&1 && { sudo -n true >/dev/null 2>&1 && echo OK || echo PASSWORD_REQUIRED; } || echo MISSING
+command -v sudo >/dev/null 2>&1 && { sudo -n true >/dev/null 2>&1 && echo OK || echo PASSWORD_REQUIRED; } || echo MISSING; echo "$D"
+command -v podman-compose >/dev/null 2>&1 && podman-compose version 2>/dev/null || echo MISSING
 """
 
 
@@ -108,6 +109,7 @@ def _parse_doctor(stdout, hostname):
     # Privilege escalation, appended so the indices above are unchanged.
     sudo_version = p(23) if len(parts) > 23 else "MISSING"
     sudo_nopasswd = p(24) if len(parts) > 24 else "MISSING"
+    podman_compose_version = p(25) if len(parts) > 25 else "MISSING"
 
     lines = [
         "Canasta Dependency Check (%s)" % hostname,
@@ -186,6 +188,37 @@ def _parse_doctor(stdout, hostname):
         else:
             lines.append(
                 "  Podman:          %s (version unknown)" % ver
+            )
+
+    if podman_compose_version not in ("MISSING", ""):
+        pc_ver = podman_compose_version
+        # podman-compose version prints two lines:
+        #   "podman version 5.4.2"
+        #   "podman-compose version 1.3.0"
+        # We must match the *second* line, not the first.
+        m = re.search(r'podman-compose version\s+(\d+\.\d+\.\d+)', pc_ver)
+        if m:
+            pc_ver_str = m.group(1)
+            pc_ver_parts = [int(x) for x in pc_ver_str.split(".")]
+            while len(pc_ver_parts) < 3:
+                pc_ver_parts.append(0)
+            if pc_ver_parts == [1, 3, 0]:
+                lines.append(
+                    "  Podman Compose:  %s — WARNING: podman-compose 1.3.0 "
+                    "has a known bug where ${VAR:-default} is passed "
+                    "literally to containers when VAR is unset. This "
+                    "breaks Canasta's docker-compose.yml defaults. "
+                    "Upgrade to >= 1.3.1, or install via: canasta "
+                    "install uv:podman-compose -H <host>"
+                    % pc_ver_str
+                )
+            else:
+                lines.append(
+                    "  Podman Compose:  %s (OK)" % pc_ver_str
+                )
+        else:
+            lines.append(
+                "  Podman Compose:  %s" % pc_ver
             )
 
     lines.append("")


cicalese (commented):
Reviewed against a checkout of the branch; full unit suite is green (2457 passed). The field-index append is done correctly — new field 25, nothing above it shifts — and the `command -v` guard means the "command not found" leak that `tests/unit/test_doctor_podman_probe.py` exists to catch can't happen through this probe.

A few things:

### The remedy it prints doesn't exist yet

The warning tells the operator to run:

```
canasta install uv:podman-compose -H <host>
```

The `uv:` target lands in #1437, which is still open. On `main` the `install` choices are `[docker, k8s-cp, k8s-worker, git-crypt, sops, podman, canasta]` (`meta/command_definitions.yml:868`), so argparse rejects that command outright. If this merges first, doctor hands out a remedy that errors.

Either gate this on #1437 landing, or soften the text to name the version requirement without the exact command.

Separately, the `-H <host>` suffix reads oddly for a local `canasta doctor` run — doctor already knows which host it probed.

### No tests

`tests/unit/test_doctor_podman_probe.py` exists specifically to hold this probe honest, and the new branch is a version-string parser with no coverage. Four cases would cover it: `1.3.0` → warning, `1.5.0` → OK, `MISSING` → no line at all, unparsable → fallback.

### The comment about output order is backwards

`podman-compose version` prints its own version first, then shells out to `podman --version`:

```
podman-compose version 1.3.0
podman version 5.4.2
```

The comment says the reverse and instructs the reader to "match the *second* line, not the first." The regex is anchored on `podman-compose version` so it works either way, but the comment will mislead whoever touches this next.

### Everything that isn't exactly 1.3.0 reports `(OK)`

Confirmed by feeding synthetic probe output through `_parse_doctor`:

```
'podman-compose version 1.0.6\npodman version 4.9.3'  ->  '  Podman Compose:  1.0.6 (OK)'
```

The podman check immediately above warns below 4.4. If there's a known podman-compose floor, mirror that; otherwise `(OK)` overstates what's been checked.

`1.3.0.post1` also matches and warns — probably the behavior you want, but it falls out of the regex rather than being deliberate, so worth a comment either way.

### Dead code

```python
pc_ver_parts = [int(x) for x in pc_ver_str.split(".")]
while len(pc_ver_parts) < 3:
    pc_ver_parts.append(0)
if pc_ver_parts == [1, 3, 0]:
```

The regex requires three components, so the padding loop can't fire and the whole comparison is `pc_ver_str == "1.3.0"`. Copied from the podman block above, where the padding is equally dead — so this is pre-existing style rather than something this PR introduced.

### Minor

The unparsable-output branch does `"  Podman Compose:  %s" % pc_ver` with a value that may be two lines, which breaks the alignment of the report. Taking the first line would be safer.

hexmode:
Addressed the review in `f8b34da`; CI is green (Unit Tests, Lint, Helm Chart Lint, Docker Smoke Test).

- **Remedy softened** — the warning no longer prints `canasta install uv:podman-compose -H <host>` (the `uv:` choice only exists once #1437 lands, and `-H <host>` read oddly for a local doctor run). It now names the requirement: `Upgrade to podman-compose >= 1.3.1` with the upstream issue link.
- **Tests added** — `test_doctor_podman_probe.py` now covers the version-string branch: `1.3.0` → warning, `1.5.0` → neutral report, `MISSING` → no line at all, unparsable multi-line → first-line fallback (aligned). `1.3.0.post1` warns too — the regex picks `1.3.0` out of it and the bug is in the base release, so it's flagged deliberately (commented).
- **Comment order fixed** — podman-compose prints its own version first, then shells out to podman; the comment now says so.
- **`(OK)` dropped** — no known podman-compose floor, so the neutral branch just prints the version rather than overstating what was checked.
- **Dead code removed** — the component-padding loop (regex already requires `\d+\.\d+\.\d+`), replaced with a direct string comparison.
- **Alignment** — the unparsable branch now takes `splitlines()[0]`.

PR #1437, opened 2026-08-04
feat(install): add uv target to install uv + podman-compose via Astral uv

### LINKED ISSUE #1435: Add  target + suggest  as fix for Debian 13's buggy podman-compose 1.3.0

hexmode:
**Problem**

Debian 13 (trixie) stable ships podman-compose 1.3.0 which has a critical bug where the `${VAR:-default}` variable-substitution syntax is passed **literally** to containers when the source and destination variable names match (containers/podman-compose#1105). This breaks Canasta's `docker-compose.yml` defaults for `MYSQL_HOST`, `PHP_MAX_INPUT_VARS`, and others.

**What this PR does**

Adds a new `canasta install uv` target that:

1. Installs the Astral `uv` Python package manager via the official standalone installer (`curl -LsSf https://astral.sh/uv/install.sh | sh`)
2. Uses `uv tool install podman-compose` to install a fixed version of podman-compose (>= 1.5.0) outside the system package manager

**Files changed**

- `roles/install/tasks/uv.yml` — new Ansible role task that performs the install
- `playbooks/install.yml` — adds `uv` to validation and dispatch
- `meta/command_definitions.yml` — adds `uv` to choices, examples, and long description

**Usage**

```bash
canasta install uv -H nichework.com
# or for localhost:
canasta install uv
```

**Related issues**

- Fixes #1435
- See also #1434 (doctor warning for podman-compose 1.3.0) and #1436 (the PR that adds the warning)

### DIFF
--- a/canasta.py
+++ b/canasta.py
@@ -722,6 +722,28 @@ def hoist_flags_from_remainder(args):
         setattr(args, pos_name, new_args)
 
 
+class _DynamicChoices:
+    """argparse choices that accept a fixed list plus a prefix pattern.
+
+    Used for `canasta install` where the base packages are fixed
+    (docker, podman, etc.) but `uv:<tool>` is an open-ended prefix
+    (e.g. `uv:podman-compose`).  argparse calls `__contains__` for
+    validation and `__iter__` for --help output.
+    """
+
+    def __init__(self, base, prefix):
+        self._base = set(base)
+        self._prefix = prefix
+
+    def __contains__(self, item):
+        if item in self._base:
+            return True
+        return isinstance(item, str) and item.startswith(self._prefix)
+
+    def __iter__(self):
+        return iter(sorted(self._base))
+
+
 def _positional_choices(param, name):
     """argparse kwargs constraining a positional to a fixed set of values.
 
@@ -732,6 +754,9 @@ def _positional_choices(param, name):
     """
     choices = param.get("choices")
     if choices:
+        prefix = param.get("choices_dynamic_prefix")
+        if prefix:
+            return {"choices": _DynamicChoices(choices, prefix)}
         return {"choices": choices}
     return {"metavar": name.upper()}
 
--- a/meta/command_definitions.yml
+++ b/meta/command_definitions.yml
@@ -775,7 +775,8 @@ commands:
     description: "Install dependencies on the target host"
     long_description: |
       Install one or more dependencies on the target host. Supported
-      packages: docker, k8s-cp, k8s-worker, git-crypt, sops, canasta, podman.
+      packages: docker, k8s-cp, k8s-worker, git-crypt, sops, canasta,
+      podman, uv, uv:<tool>.
 
       `k8s-cp` installs k3s as a control-plane node — the cluster's
       API server, scheduler, etc., plus kubectl + helm on the same
@@ -823,6 +824,13 @@ commands:
       rootless Podman targets. Idempotent — skips if podman is already
       installed.
 
+      `uv` installs the Astral uv Python package manager on its own.
+      `uv:<tool>` installs `<tool>` via uv (e.g. `uv:podman-compose`).
+      This is the recommended workaround for Debian 13 (trixie) stable,
+      which ships podman-compose 1.3.0 — a version with a known bug
+      that breaks Canasta's docker-compose.yml defaults. Idempotent —
+      skips if the tool is already present.
+
       Multiple packages can be combined in a single command, except
       that `k8s-cp` and `k8s-worker` are mutually exclusive on a given
       host.
@@ -837,6 +845,10 @@ commands:
       - "canasta install -i mysite canasta"
       - "canasta install podman"
       - "canasta install --host prod1.example.com podman"
+      - "canasta install uv"
+      - "canasta install --host prod1.example.com uv"
+      - "canasta install uv:podman-compose"
+      - "canasta install --host prod1.example.com uv:podman-compose"
     playbook: install.yml
     parameters:
       - name: host
@@ -863,10 +875,11 @@ commands:
         positional: true
         multi: true
         required: true
-        choices: [docker, k8s-cp, k8s-worker, git-crypt, sops, podman, canasta]
+        choices: [docker, k8s-cp, k8s-worker, git-crypt, sops, podman, canasta, uv]
+        choices_dynamic_prefix: "uv:"
         description: >-
           Dependencies to install (docker, k8s-cp, k8s-worker,
-          git-crypt, sops, podman, canasta)
+          git-crypt, sops, podman, canasta, uv, uv:<tool>)
       - name: public_ip
         type: string
         multi: true
--- a/playbooks/install.yml
+++ b/playbooks/install.yml
@@ -1,7 +1,7 @@
 ---
 # canasta install - Install dependencies on the target host.
 # Accepts one or more package names: docker, k8s-cp, k8s-worker,
-# git-crypt, sops, canasta, podman.
+# git-crypt, sops, canasta, podman, uv, uv:<tool>.
 #
 # Package names are the CLI's own, dispatched on as-is. The task files
 # below keep the name of the software they install (k3s), which is not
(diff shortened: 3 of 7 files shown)


cicalese (commented):
Reviewed against a checkout of the branch. The `_DynamicChoices` idea is sound and the playbook wiring is the right shape, but CI is red on **Unit Tests** as well as Lint, `make validate` fails locally, and there's a design problem that may mean the install never takes effect.

## The functional problem: the uv copy is shadowed

`uv tool install podman-compose` puts the binary in `~/.local/bin`, and this PR tries to reach it by appending to `~/.profile` and `~/.bashrc`. But Canasta runs compose as a bare name through the command module (`roles/orchestrator/tasks/start.yml:88`):

```yaml
ansible.builtin.command:
  cmd: "{{ _compose_cmd }} {{ _compose_files | join(' ') }} ... up -d"
```

That resolves against the PATH of a non-interactive, non-login SSH session. `~/.profile` is read only by login shells, and Debian's `~/.bashrc` early-returns when non-interactive — so neither file is consulted.

Worse: `roles/install/tasks/podman.yml:81` installs `podman-compose` from apt into `/usr/bin`. On the exact host this PR targets, `command -v podman-compose` finds the broken 1.3.0 and the uv copy is shadowed. The same applies to the doctor probe in #1436, so the two PRs won't agree on what's installed.

The hook that already exists for this is `compose_command` — the registry/`.env` override that `roles/common/tasks/detect_runtime.yml:55` sets. Pointing it at the absolute `~/.local/bin/podman-compose`, or symlinking into `/usr/local/bin` (which precedes `/usr/bin` on Debian's default PATH), would actually take effect.

I verified the invocation mechanism and the rc-file behavior, not an end-to-end run on trixie — but the shadowing looks certain.

## CI failures

**`test_install_rejects_malformed_uv_prefix`** (this PR's own new test). `_DynamicChoices.__contains__` returns True for anything starting with `uv:`, including bare `uv:`. So `canasta install uv:` clears argparse and is only caught by the playbook's `^uv:[a-zA-Z0-9_-]+$` regex — after Ansible has started and SSHed to the target. The test is right; `__contains__` needs to require a non-empty tail.

**`make validate` fails with 10 errors.** Run locally:

```
- commands[11] (install).parameters[3] (packages): 'canasta' is an accepted choice but nothing dispatches on it
  ... same for docker, git-crypt, k8s-cp, k8s-worker, podman, sops, uv
- commands[11] (install): example 'canasta install uv:podman-compose': packages: 'uv:podman-compose' is not one of docker, k8s-cp, ...
- commands[11] (install): example 'canasta install --host <host> uv:podman-compose': (same)
```

Two causes. The `_install_packages` → `_regular_packages` rename breaks `check_dispatch` in `scripts/validate_definitions.py`, which greps for `'<pkg>' in _install_packages`; the PR patched the unit-test copy of that check in `test_install_podman.py` to accept either name but left the validator alone. And the example checker validates against the static `choices` list, which knows nothing about `choices_dynamic_prefix`.

(The green "validate" check on this PR came from a different workflow run than the failing one — worth not trusting.)

**`test_gitops_sops_scaffold.py::test_install_supports_sops_toolchain`** and **`test_install_choices_and_branches_agree`** both fall out of the same rename.

### The rename looks unnecessary

`'docker' in _install_packages` is unaffected by a `uv:podman-compose` element sitting in the list — `uv:<tool>` never equals a base package name, and `'uv' in _install_packages` is equally correct since `uv != uv:x`. Dropping the rename removes 8 changed dispatch lines and clears four of the five test failures. `_regular_packages` is still useful as an input to the uv role, just not as the dispatch variable.

## Bugs in `roles/install/tasks/uv.yml`

**Clobbering `ansible_env`:**

```yaml
- name: Ensure ~/.local/bin is on PATH for this session
  ansible.builtin.set_fact:
    ansible_env:
      PATH: "{{ _install_home }}/.local/bin:{{ ansible_env.PATH | default(...) }}"
```

This does two wrong things. It doesn't work — `ansible_env` is a gathered fact; task environments come from `environment:`, not from overwriting it. And it's destructive: confirmed with a local playbook that after this `set_fact`, `ansible_env` contains only `PATH` for the rest of the play. `HOME` is gone, and `roles/install/tasks/uninstall_k3s.yml:164` and `roles/orchestrator/tasks/k8s_install_k3s.yml:166` both read `ansible_env.HOME`. Different plays today, so probably not reachable — but it's a landmine. The task is also unused: every subsequent task uses the absolute `_uv_bin`.

**The PATH `when:` evaluates on the wrong machine:**

```yaml
when: "_install_home ~ '/.local/bin' not in lookup('env', 'PATH') | default('')"
```

`lookup('env', ...)` runs on the controller. This gates a file edit on the *target* against the *controller's* PATH.

**Unreachable error handler.** `Warn if uv installer failed` has `when: _uv_install is failed`, but the shell task above it has no `ignore_errors`, so a failure aborts the block before the debug task runs.

**Failures are swallowed.** `ignore_errors: true` on the tool install means `canasta install uv:podman-compose` exits 0 having installed nothing, with only a `debug` WARNING in the output. That's the opposite of how the rest of the repo handles install failures.

**No version pin.** `uv tool install podman-compose` takes whatever PyPI serves. The doctor message in #1436 says ">= 1.3.1" — pinning it here would keep a stale index or a regression from reinstalling the broken version.

## Smaller

- `_install_user` is computed and never used.
- `creates: "{{ _uv_bin }}"` duplicates the `_uv_stat` guard, and `changed_when: true` reports changed unconditionally.
- Writing to both `.profile` and `.bashrc` with near-identical tasks — pick one, or drop both if you switch to the `compose_command` approach.
- No tests for `uv.yml` itself; `tests/unit/test_install_podman.py` is the model.
- `curl -LsSf https://astral.sh/uv/install.sh | sh` is unpinned and unverified. There's precedent (the k3s installer), so probably fine, but worth a comment saying so.
- The known lint failure is `risky-shell-pipe` at `uv.yml:46` — needs `set -o pipefail`.
- `_DynamicChoices.__iter__` returns `sorted(self._base)`, so `--help` lists packages alphabetically rather than in definition order, unlike every other command.

The blocking work is the PATH/shadowing question, then reverting the rename and tightening the `uv:` validation.

PR #1439, opened 2026-08-04
fix(docker): let 'canasta install' target a remote host from docker mode

### LINKED ISSUE #1438: Allow 'canasta install <pkg> -i <instance>' from canasta-docker by bootstrapping canasta-native on the target
## Problem

`canasta install` from a canasta-docker controller is rejected outright, even when it names a target:

```
$ canasta install uv:podman-compose -i nichework
Error: 'canasta install' is not supported in canasta-docker mode.
...
```

The `-i`/`-H` options exist for `canasta install` (install on the host where the instance is registered), but the canasta-docker wrapper's guard (#535) rejects **every** install invocation before the CLI even parses them.

## Why a remote install can work from docker mode

The guard's rationale is that installs provision software on the HOST (k3s as a systemd unit, docker engine, etc.) which can't run from inside the canasta-ansible container. That's true only for a **local** install, where 'localhost' is the container. For a **remote** target the install playbook already runs over SSH on the target with root — exactly what `canasta install -H <host> <package>` does in native mode, and exactly how every other remote `-H` command works from docker mode.

## Requested behavior

- `canasta install <package> -i <instance>` (or `-H <host>`) from canasta-docker: resolve the target (`-i` via the registry) and pass the command through to the container, which runs the install playbook over SSH on the target.
- Keep the rejection only for genuinely-unsupported local installs (no target, or a target resolving to the controller itself).

## Use case

A canasta-docker controller managing a remote Podman host running Debian 13 (trixie), which ships podman-compose 1.3.0 — the version with the broken ${VAR:-default} substitution (containers/podman-compose#1105). The workaround is `canasta install uv:podman-compose`, which today is impossible from docker mode; the operator must SSH in manually.

hexmode:
## Summary

`canasta install` was unusable from canasta-docker: the wrapper rejected **every** invocation with `Error: 'canasta install' is not supported in canasta-docker mode` (#535), even when the command named a remote target via `-i <instance>` or `-H <host>`.

The rejection is correct for a **local** install — inside the container there's no systemd and no host package manager — but for a **remote** target the install playbook already runs over SSH on the target with root, exactly like every other remote `-H` command in docker mode. This PR lets remote-targeted installs through.

## Behavior

```
$ canasta install uv:podman-compose -i nichework          # now: runs over SSH on nichework.com
$ canasta install podman -H prod1.example.com            # now: runs over SSH on prod1.example.com
$ canasta install docker                                 # still refused (no target)
$ canasta install docker -i local-instance               # still refused (target resolves to localhost)
$ canasta install sops -i unknown                        # refused with 'instance not registered'
```

### canasta-docker

Resolve the install target from `-H`/`--host` or, for a bare `-i`/`--id`, from the instance's registered host in the registry. When the target is missing or is the controller itself (localhost/127.0.0.1), keep the existing rejection — with an added hint for the `-i` case and a pointer at the now-working remote form. A remote target passes the command through to the container unchanged; the CLI resolves `-i` to the host and the playbook runs over SSH.

### canasta.py

Exempt `install` from the docker-mode Podman guard (`check_docker_mode_can_reach_runtime`). Install never drives an instance's compose stack, and in docker mode the wrapper only lets it through when it targets a remote host — so a local Podman instance registered in the registry is irrelevant and must not block a `-H` install.

## Testing

- `tests/unit/test_canasta_docker.py`: new `TestInstallDockerMode` covers pass-through for `-i`/`-H`/long-form targets, preservation of extra flags (`--public-ip`), and the retained refusals (no target, unknown instance, localhost instance, `-H localhost`).
- Full unit suite: 2464 passed, 2 skipped.

Closes #1438.

### DIFF
--- a/canasta-docker
+++ b/canasta-docker
@@ -24,30 +24,6 @@ set -euo pipefail
 CANASTA_CLI_BIN="$(readlink -f "${BASH_SOURCE[0]}" 2>/dev/null || echo "${BASH_SOURCE[0]}")"
 export CANASTA_CLI_BIN
 
-# `canasta install` provisions software ON THE HOST — k3s as a systemd
-# unit, the docker engine package, etc. That work can't run inside the
-# canasta-ansible container, which has no systemd and no privilege over
-# the host package manager. Reject these commands early with a clear
-# pointer to the supported alternatives, instead of failing later with
-# a cryptic "Can not find systemd" from the k3s installer.
-if [[ "${1:-}" == "install" ]]; then
-    cat >&2 <<'INSTALL_GUARD_EOF'
-Error: 'canasta install' is not supported in canasta-docker mode.
-
-These commands install system packages on the host (k3s as a systemd
-unit, docker engine, etc.) which can't run from inside a container.
-
-Either:
-  - Use canasta-native:
-      curl -fsSL https://get.canasta.wiki | bash -s -- --native
-  - Or install the dependency manually on the host (e.g.
-      curl -sfL https://get.k3s.io | sh -
-    for k3s) and use canasta-docker for the rest of the workflow
-    (create, start, backup, restore, etc.).
-INSTALL_GUARD_EOF
-    exit 1
-fi
-
 # CANASTA_CLI_IMAGE is the post-rename name; CANASTA_ANSIBLE_IMAGE
 # remains a backwards-compatible fallback for users who already
 # set the old name in their environment. When either is set the user has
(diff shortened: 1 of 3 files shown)


cicalese (commented):
Reviewed against a checkout of the branch; full unit suite is green (2466 passed). The premise is right, and the `canasta.py` exemption is genuinely needed — with no `-i`, `check_docker_mode_can_reach_runtime` (`canasta.py:945`) scans *every* registered instance, so one local Podman instance would otherwise block an unrelated `-H` install. Pass-through works for `-H`, `--host=`, `-i` and `--id=`.

Three reproducible bugs, all in the wrapper's target resolution.

### A locally-created instance is reported as "not registered"

`instance_to_dict` (`roles/common/library/canasta_registry.py:155`) writes the `host` key only when it's truthy:

```python
if instance.get("host"):
    result["host"] = instance["host"]
```

So an instance created without `--host` has **no** `host` key at all. Reproduced with a registry entry of the shape `canasta create` actually writes:

```
$ cat conf.json
{"Instances": {"mysite": {"id": "mysite", "path": "/srv/mysite", "orchestrator": "compose"}}}

$ canasta install docker -i mysite
Error: instance 'mysite' is not registered, so there is no
host to install on. Register it with 'canasta create' or 'canasta host add'.
```

The instance *is* registered, and the message tells the operator to create it again. It also loses the useful docker-mode explanation the other branch prints.

`test_install_by_i_for_a_local_instance_is_refused` seeds `"host": "localhost"` explicitly — a shape the registry never produces — so the test passes while the common case is broken. Missing or empty `host` should mean *local*; "not registered" should be reserved for an ID that isn't in `Instances` at all.

### Attached short options are refused

argparse accepts `-Hprod1.example.com` and `-imysite`; the wrapper's scan handles only the separated and `=` forms. Both of these name a legitimate remote target and both get refused:

```
$ canasta install podman -Hprod1.example.com
Error: 'canasta install' is not supported in canasta-docker mode.
```

### Any pre-subcommand global flag bypasses the guard entirely

The scan is gated on `${1:-} == "install"`, but `canasta.py` explicitly supports pre-command args (`canasta.py:1760-1795`), and `--host <h> start -i <id>` is the documented form. So both of these sail straight through in docker mode:

```
canasta --host localhost install podman
canasta -v install docker
```

That contradicts the behavior table in the PR description (`-H localhost` still refused). This was true of the old guard too, but this is the PR that turns the guard into a target decision, so it's the natural place to fix it — find the `install` token among the leading flags rather than testing `$1`.

### Design: the resolution probably belongs in canasta.py

All three of the above fall out of doing registry resolution in bash. The wrapper re-implements the lookup (plus a `grep -A15` fallback), and `canasta.py` then has to blindly trust the result — the exemption is a blanket `command_name != "install"`, so invoking `canasta.py` directly with `CANASTA_RUN_MODE=docker` gets no guard at all.

Python already parses the registry properly and argparse has already resolved `-i`/`-H`. Moving the decision there fixes all three bugs for free and drops ~60 lines of bash.

If the reason for keeping it in the wrapper is failing before the container starts, that's a legitimate trade — but it should be stated in the comment, because it isn't obvious from the code.

### Smaller

- `'$INSTALL_INSTANCE'` and `'$CONFIG_DIR'` are interpolated into Python source; a quote in either breaks the lookup. Consistent with the existing lookup at `canasta-docker:360`, but this one gates the allow/refuse decision, so passing them via `sys.argv` would be worth it here.
- The `grep -A15` fallback can pick up a *later* instance's `host` when the named one has none — which, after the first fix above, is the common shape.
- `test_non_install_commands_are_unaffected` asserts `"install" not in user_args(argv)` for `["version"]`; that holds no matter what the guard does.

hexmode:
Addressed review (2026-08-05): the install-target decision moved out of the bash wrapper into canasta.py (`check_docker_mode_install_target`), fixing all three wrapper bugs:

1. **Missing `host` key** — `canasta create` writes no `host` key for a locally created instance, so the wrapper read `inst.get("host", "")` and wrongly said "not registered". The registry lookup now uses the real shape: a present ID with a missing/empty/local host is a *local* install (refused with the docker-mode explanation); "not registered" is reserved for an ID absent from `Instances`.
2. **Attached short options** — argparse handles `-Hprod1.example.com` / `-imysite`, which the wrapper's separated `-H`/`-i` scan missed.
3. **Pre-command flags** — `canasta -v install docker` puts `install` in \$2, so the wrapper's \$1==install guard never fired and the install playbook actually ran on the controller from inside the container. The guard now runs in main() for the install command regardless of where flags appear.

Trade-off acknowledged in code: refusal now happens after the container starts rather than before (the wrapper passes everything through). That buys correctness for the three cases above.

Also per the review, `--host` wins over `--id` when both are given (install already documents this), and `-H` on a localhost/127.0.0.1 value refuses regardless of `-i`.

Tests: new `tests/unit/test_docker_mode_install_guard.py` (unit guard tests, argparse-level attached/pre-command forms, and end-to-end canasta.py runs), and `TestInstallDockerMode` now asserts the wrapper passes everything through. CI green (Lint, Unit Tests, Helm Chart Lint, Docker Smoke Test; integrations skipped pending approval).

canasta-docker
```diff
@@ -70,6 +46,20 @@ fi
...
+# where the install playbook runs over SSH and systemd and root exist.
+#
+# The wrapper passes install through unchanged and lets canasta.py
+# refuse local installs. That keeps the decision where the args have
+# already been parsed: argparse handles the attached short forms
+# (-Hprod1.example.com, -imysite) and pre-command flags ('canasta -v
+# install docker') that a bash scan of "$@" would miss, and the -i
+# lookup reads the real registry (an instance created without --host has
+# no 'host' key at all, which means local). See
+# check_docker_mode_install_target in canasta.py.
```
cicalese, line 61: Worth stating here that this moves the refusal inside the container: a local `canasta install` now pulls and starts the CLI image before printing the error, where the old wrapper guard failed instantly. Reasonable trade for getting the target right, but it isn't obvious from the comment as written.

canasta.py
```diff
@@ -1011,6 +1011,86 @@ def check_docker_mode_can_reach_runtime(args):
...
+    if os.environ.get("CANASTA_RUN_MODE") != "docker":
+        return
+
+    host = getattr(args, "host", None)
+    inst_id = getattr(args, "id", None)
+    if host:
+        # -H/--host names the controller itself: local, refuse. Any other
+        # host is remote and passes through. --host wins over --id if both
+        # are given, so a -H here settles it without reading the registry.
+        if host in ("localhost", "127.0.0.1"):
```
cicalese: This introduces a second definition of "local": `_helpers._is_localhost` is `host in ("localhost", "", None)` and notably excludes `127.0.0.1`. This one is the better definition — but two of them will drift, so consider centralizing on it.

cicalese (commented):
Re-reviewed after the rework. All three bugs from the last round are fixed, each with a named regression test, and CI is green on this head. Two small things.

cicalese:
Pushed a commit addressing both review comments, so this can go into the upcoming release.

**One definition of "local".** The guard carried its own `("localhost", "127.0.0.1")` list while `_helpers._is_localhost` said `("localhost", "", None)`. Added `_is_local_target` beside `_is_localhost` and pointed the guard at it.

Deliberately a separate helper rather than widening `_is_localhost`: that one answers "do we need ssh to reach this path?" for roughly thirty call sites, and adding the loopback literals there would silently reroute an instance registered as `127.0.0.1` from ssh to local execution. The new helper answers the narrower question the guard asks, and picks up `::1`, which neither definition had. A test asserts `_is_localhost` was left alone, so the wider change cannot arrive later by accident.

**The trade is now recorded.** The wrapper comment explained where the decision moved and why, but not what it costs: the refusal happens inside the container now, so a local `canasta install` pulls and starts the CLI image before printing the error, where the wrapper-side guard failed instantly.

Four tests added; 2633 unit tests, `make lint`, `make validate` green.

